PR #2, opened 2023-03-14
Add initial structure

paddyroddy:


p-j-smith (approved):
nice! Do we need to specify `version` here, or could we use dynamic versioning the pyproject.toml? No need to change it now though

paddyroddy:
This guy has one https://github.com/audreyfeldroy/cookiecutter-pypackage but I think we can remove at a later date if we can

PR #9, opened 2023-03-14
Clearer author fields

dstansby:


cookiecutter.json
```diff
@@ -1,8 +1,8 @@
 {
-  "full_name": "UCL ARC",
-  "email": "",
-  "github_username": "{{ cookiecutter.full_name.replace(' ', '-') }}",
-  "project_name": "Python Template",
+  "author_name": "",
+  "author_email": "",
+  "github_username": "{{ cookiecutter.author_name.replace(' ', '-') }}",
+  "project_name": "",
```
p-j-smith: ah, CI is failing because there's a test to check that the folder `python-template` exists. Maybe keep the default as "Python template"? Or change the test if you prefer

samcunliffe:
I updated so this should be good on the current `main`.

paddyroddy (approved):
Looks good. Although, @samcunliffe said something about wanting a first name/surname thing?

samcunliffe:
> Looks good. Although, @samcunliffe said something about wanting a first name/surname thing?

Na, I'm good. I can split the `author_name`.

PR #25, opened 2023-05-02
13 how to integrate setuptools scm

### LINKED ISSUE #13: How to integrate `setuptools-scm`?
Ideally, we'd have something like the following included in the `pyproject.toml`.

```toml
[build-system]
requires = [
    "setuptools-scm",
]

[project]
dynamic = [
    "version",
]

[tool.setuptools_scm]
```

This has the advantage that all files tracked by `git` are included in the build - avoiding the need for `MANIFEST.in` etc. Also, we can do dynamic versioning. 

If we have this, I run into issues as the templated generated code isn't a `git` repo. How can we get this to work nicely?

ruaridhg:
Trying to get dynamic versioning to work


paddyroddy (approved):
Yay, thanks for looking into this!

PR #46, opened 2023-05-18
Just require minimum python version

paddyroddy:
This way a user is free to use `3.12` for example but it's just that the package officially only supports up to, e.g. `3.11`.

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -44,7 +44,7 @@ optional-dependencies = {dev = [
     "twine",
 ]}
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


samcunliffe:
Should we remove the option from the cookie config? Or add a "no max version" default?

samcunliffe (approved):

paddyroddy:
> Should we remove the option from the cookie config? Or add a "no max version" default?

No! Look through the code. A lot of nice stuff is done from the max. Happy to discuss, but the only thing that needs changing is this. Happy if you want, we could just set the max value according to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

paddyroddy:
e.g.

`classifiers` for `PyPI`
image

`tox`
image

tests in CI
image

samcunliffe:
> No! Look through the code.

Gottit.

PR #66, opened 2023-07-14
Pin `tox>=4`

### LINKED ISSUE #64: tox configuration requires tox version 4
While using this template for https://github.com/UCL/dxh (thanks all for the work in getting this set up 😁) I found that the default `legacy_tox_ini` in the `[tool.tox]` section of the generated `pyproject.toml` 

https://github.com/UCL-ARC/python-template/blob/ddf6cc2f8388efa9b9b2c8873f2ea5e837247b01/%7B%7Bcookiecutter.project_slug%7D%7D/pyproject.toml#L144-L169

needed a few tweaks before I could get `tox` running locally with the version I had installed (v3.24.5)

- Removing the whitespace indentation within the `""" .... """` multiline string
- [Adding `isolated_build = true` to the `[tox]` section](https://tox.wiki/en/3.24.5/config.html#conf-isolated_build)

Might be worth documenting this and/or pinning the `tox` version in the `dev` list to `tox>=4` in `optional-dependencies` . Potentially may also be worth adding version pin when `tox` is installed in Actions workflows though not sure if it is likely that a previous version will get installed on Actions runners in practice.

paddyroddy:
Fixes #64

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -36,7 +36,7 @@ optional-dependencies = {dev = [
     "mypy",
     "pre-commit",
     "ruff",
-    "tox",
+    "tox>=4",
     "twine",
 ]}
 readme = "README.md"


samcunliffe (approved):
Ah. Easy peasy.

Thanks for the explanation about `legacy_tox_ini`. TIL.

PR #128, opened 2023-07-31
Hyphens vs underscores: the bug reported in #127.

### LINKED ISSUE #127: Improving default for `project_slug`
I think the default for `project_slug` could possibly be improved - specifically I'm not sure replacing spaces and underscores in `project_name` with hyphens gives a good model of what this should be as to be a valid package or module name we cannot have hyphens and to align with [PEP8 recommendations](https://peps.python.org/pep-0008/#package-and-module-names) we shouldn't have underscores in the package name either.

_Originally posted by @matt-graham in https://github.com/UCL-ARC/python-tooling/issues/108#issuecomment-1652250241_

Somewhat relatedly I wonder if we should also add a pre-generate hook to check whether the `project_slug` chosen conflicts with an existing package (distribution) on PyPI (probably just warning the user rather than preventing this being possible).

samcunliffe:
Cheers @matt-graham. Good spot.


cookiecutter.json
```diff
@@ -4,6 +4,7 @@
   "github_username": "{{ cookiecutter.author_name.replace(' ', '-') }}",
   "project_name": "Python Template",
   "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '-').replace('_', '-') }}",
+  "module_name": "{{ cookiecutter.project_slug.replace('-', '_') }}",
```
matt-graham: Should this be `module_name` or `package_name`? This gets to a wider confusion I have with how the template is set up. At the moment if `module_name` is set to `test_project` the project directory will be intialised as

```tree
.
├── LICENCE.md
├── README.md
├── pyproject.toml
├── src
│   ├── test_project
│   │   ├── _version.py
│   │   └── test_project.py
└── tests
    └── test_dummy.py
```

If `src/test_project` contained an `__init__.py` file this I think would correspond to there being an importable package `test_project` which contains a module with name `test_project` and so we could import both `test_project` and `test_project.test_project` (with the former bringing in any definitions in the `src/test_project/__init__.py` files and the latter in `src/test_project/test_project.py`). This would correspond to the example at  `src` layout example at  https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

The template however does not create an `__init__.py` file by default. I wondered if this meant instead the template is intended to correspond to a single-module distribution without an enclosing package. However, if I install an instantiation of the template corresponding to the above directory tree then both `import test_project` and `import test_project.test_project` run, with `test_project.test_project.__doc__`  evaluating to `'An example file.'` (the module docstring contained in `src/test_project/test_project.py`. This suggests the outer `test_project` directory is being identified as a package despite the lack of `__init__.py` file. 

It seems that [the support for implicit namespace packages from Python 3.3 does allow for packages to be imported without an `__init__.py` file being present](https://stackoverflow.com/a/37975003) so I think this might be what is causing the directory tree above to be importable as a package - support this is that if I run `import test_project` then `test_project.__path__` returns `_NamespacePath(['.../envs/test-project/lib/python3.11/site-packages/test_project'])` and  `test_project.__file__` is empty which seems to [suggest that test_project is a namespace rather than regular package](https://peps.python.org/pep-0420/#differences-between-namespace-packages-and-regular-packages).

tests/conftest.py
```diff
@@ -15,5 +15,7 @@ def project_config():
         as defined in the cookiecutter.json
     """
     return {
-        "project_slug": "cookiecutter_test",
+        "project_name": "Cookiecutter Test",
+        "expected_repo_name": "cookiecutter-test",
+        "expected_slug": "cookiecutter_test",
```
matt-graham: I would say `expected_module_name` (or `expected_package_name` depending on resolution of earlier comment!) would better reflect what we are testing for here (as I think the tests which use this are checking for files / directories for which the name should be set by `module_name` not `project_slug`).

cookiecutter.json
matt-graham: Re-reading that comment I realised that was a lot of text without a concrete question / query 😅. Main question is do we want the template to create a package or single module distribution? If the former then I think `package_name` would be more appropriate and we should probably include a (possibly empty) `__init__.py` so we get a regular rather than namespace package. If we want to create a single module distribution then `module_name` is fine but we should create the module directly in the `src` directory.

EDIT: Correct shouldn't to should in final sentence.

samcunliffe: I think I softly incline towards: `test-project/src/test_project/__init__.py` and _no_ `test-project/src/test_project/test_project.py` file. (The latter has funny imports..).

Let's see what @paddyroddy says.

p-j-smith: > I think I softly incline towards: test-project/src/test_project/__init__.py and no test-project/src/test_project/test_project.py file. (The latter has funny imports..).

I think having `test-project/src/test_project/__init__.py ` definitely makes sense. But I find having no modules a little strange. Perhaps we could call it `test-project/src/test_project/main.py` instead? Or have both `package_name` and `module_name` options, where `module_name` defaults to `main`?

paddyroddy: The idea behind setting this up was just to make it really obvious to someone new to `python` where to put their package. The folder structure alone may not be enough. When I set the package structure up, it was never intended to be the final result 😛

We are indeed doing a `src` layout, as I've read plenty about the advantages of such an approach.

As pointed out above, one can have a package without an `__init__.py` (although it is contentious). I think the suggestion of `test-project/src/test_project/main.py` might be better, as people can then decide if they want an `__init__.py` or not.

As I mentioned before, most structure ideas (i.e. `project_slug`) have come from https://github.com/audreyfeldroy/cookiecutter-pypackage, so my thinking was to have it in a similar structure to others - but I don't really mind what we do

tests/test_package_gen.py
```diff
@@ -33,30 +33,35 @@ def test_package_generation(
...
         "README.md",
         ".pre-commit-config.yaml",
         "LICENCE.md",
         "pyproject.toml",
         "src",
-        Path("src") / project_config["project_slug"],
+        Path("src") / project_config["expected_slug"],
+        Path("src")
+        / project_config["expected_slug"]
+        / (project_config["expected_slug"] + ".py"),
```
paddyroddy:
```suggestion
        / f"{project_config['expected_slug']}.py",
```

paddyroddy (approved):

cookiecutter.json
matt-graham: My 🪙🪙 would be that I think that having the template correspond to a regular rather than implicit namespace package (and so including an `__init__.py` file) would be the better default for a new Python user as (i) regular packages are (in my experience) much more common in practice and most packaging documentation / tutorials that I've come across assume the user is creating a regular package so having our default be something which doesn't correspond to the structure recommended in such documentation may confuse users (as it confused me when trying to figure out how the package was being imported successfully) and (ii) there is a performance advantage at import time to regular packages:

> If a developer knows that her package will never be a portion of a namespace package, then there is a performance advantage to it being a regular package (with an \_\_init__.py). Creation and loading of a regular package can take place immediately when it is located along the path. With namespace packages, all entries in the path must be scanned before the package is created.
> _From https://peps.python.org/pep-0420/#discussion_

As another point in favour of including `__init__.py` https://github.com/audreyfeldroy/cookiecutter-pypackage [includes one](https://github.com/audreyfeldroy/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/__init__.py) in their template 😉

paddyroddy:
@samcunliffe what's happening with this?

samcunliffe:
@paddyroddy sorry... my bad. I let it go a bit stale.

There are some action items for me from the comments above:

- [x] ~~slug~~ module → package in variable names and everywhere
- [x] (consensus seems to be) add an `__init__.py`

samcunliffe:
Sorry for being slow.

paddyroddy:
Can you resolve conflicts before review @samcunliffe?

samcunliffe:
> Can you resolve conflicts before review @samcunliffe?

I did at the time of requesting. But there's too much development velocity here.

paddyroddy:
> I did at the time of requesting. But there's too much development velocity here.

Don't hate the player, hate the game

paddyroddy (approved):
Looks good @samcunliffe

samcunliffe:
Merging before I get overtaken by development velocity. @p-j-smith @matt-graham please shout if you want to revisit anything here.

PR #155, opened 2023-08-10
Write a coverage report when running tox

dstansby:
When setting up code coverage, I noticed there wasn't a coverage file written that codecov could use. This fixes that by writing an xml coverage file.

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -153,7 +153,7 @@ legacy_tox_ini = """
 
     [testenv]
     commands =
-        pytest --cov
+        pytest --cov --cov-report=xml
     deps =
         pytest
         pytest-cov


matt-graham (approved):
Good catch!

{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -153,7 +153,7 @@ legacy_tox_ini = """
 
     [testenv]
     commands =
-        pytest --cov
+        pytest --cov --cov-report=xml
```
paddyroddy, line 156: Why `xml` format? I read something about `lcov` which is why I've used it https://github.com/paddyroddy/python-template/blob/762dc809b5d2b6ac201d6fcee304add0bec34792/pyproject.toml#L173, but that might have been something specific to `coveralls`

dstansby: No particular reason, just that it's what I've always used and it works fine with codecov.io.

PR #156, opened 2023-08-11
Explicitly specify mypy config file in pre-commit-config.yaml

dstansby:
I've found that without this `mypy` doesn't read the configuration when pre-commit is run.


paddyroddy (approved):
Seems to be the same thing as `ruff` from above. It's to do with the two `pyproject.toml`s

paddyroddy:
Fixed linting and merging

Issue #49, opened 2023-06-05
Jupyter notebook formatting


paddyroddy:
Wonder whether we could make them better in git using this https://github.com/nbQA-dev/nbQA in `pre-commit`

p-j-smith:
I don't think we currently have any notebooks in the template? Or do you mean we should add this to the docs as a recommended tool?

paddyroddy:
I meant in the `pre-commit` of the template as our users might have notebooks.

samcunliffe:
Me and @krishnakumarg1984 (and @matt-graham ) have exactly this. We're supporting some researchers who are using Jupiter notebooks and we probably want to lint them without imposing **_too much_** of a workflow change. 

Some webpage recommendations (even if not included in the package template) would be appreciated.

We can probably repurpose this ticket for documenting what we learn. And/or add to the template pre-commit config as a final task, if we want that?

Some claims found whilst browsing:
* [`black[jupyter]` works](https://black.readthedocs.io/en/stable/getting_started.html#installation)
* [`ruff` works](https://beta.ruff.rs/docs/faq/#does-ruff-support-jupyter-notebooks), and integrates well with...
* [nbQA](https://github.com/nbQA-dev/nbQA) which also works as a [pre-commit tool](https://nbqa.readthedocs.io/en/latest/pre-commit.html#pre-commit).

samcunliffe:
@paddyroddy do you still want this in the template? Or is documenting it enough?

paddyroddy:
I would have thought `pre-commit` hooks? Worst case is adding a minor delay if not using them

dstansby:
I think `ruff` does jupyter notebook formatting now, at least I've seen it happening in other repos.

Saransh-cpp:
Yep, ruff does support formatting jupyter notebooks - https://docs.astral.sh/ruff/configuration/#jupyter-notebook-discovery

samcunliffe:
I keep forgetting: did we say we were just going to have the notebooks page? Or do we want the template to actually provide a `will_your_project_use_notebooks` variable?

paddyroddy:
> I keep forgetting: did we say we were just going to have the notebooks page? Or do we want the template to actually provide a `will_your_project_use_notebooks` variable?

Looks like there is a separate issue for this https://github.com/UCL-ARC/python-tooling/issues/258

paddyroddy:
[...]

PR #157, opened 2023-08-30
Add a Jupyter notebook page.

samcunliffe:
Putting my best linters found so far. Does not completely solve #49.

Anyone with experience linting jupyter notebooks is very welcome to review.


p-j-smith:
I've never done linting of notebooks so not much help here, but when you say:

> Does not completely solve https://github.com/UCL-ARC/python-tooling/issues/49.

do you know what else needs to be done to close the issue?

docs/pages/jupyter-notebooks.md
```diff
@@ -0,0 +1,40 @@
+---
+title: Jupyter Notebooks
+layout: default
+---
+
+# Jupyter notebooks
+
+We generally recommend developing code for packaging into python modules if possible.
+However occasionally we support researchers who prefer a notebook environment or projects which want to provide examples and tutorials in notebook format.
```
matt-graham:
```suggestion
We generally recommend packaging resuable code components into Python modules where possible.
However occasionally we support researchers who prefer a notebook environment or projects which want to provide examples and tutorials in notebook format.
Notebooks can also be a valid alternative to Python scripts for running and recording the results of numerical experiments for example.
```

I personally think of notebooks more of as an alternative to writing a script which runs some analysis but with the benefit of allowing richer Markdown notes and outputs like plots to be integrated alongside the code. Might be on my own here though!

matt-graham (approved):
Looks good to me other than I would personally caveat the recommendation to favour Python modules a bit as I feel notebooks do have valid use cases.

PR #171, opened 2023-09-29
Add trailing new line to BSD license

dstansby:
I just created a new package with a BSD license, and `pre-commit` failed on it because there wasn't a trailing new line at the end of `LICENSE.md`. I'm hoping this will fix that?


samcunliffe (approved):

samcunliffe:
Oh `prittier` is complaining!? 🫠

dstansby:
Seems like this is not the right way to fix this 😢

PR #173, opened 2023-09-30
Remove pydantic from pyproject.toml

dstansby:
Since not all projects will use `pydantic`, I think it makes sense to remove these lines from `pyproject.toml`? In a new project I started I'm getting errors complaining that mypy can't find pydantic, and I don't want to install pydantic because I don't need it for the project.


paddyroddy (approved):
Fair enough. My thought was more that is useful, and not obvious how to make this template compatible with it. But we can remove it for now

PR #180, opened 2023-10-07
Use default coverage

paddyroddy:
Turns out you can have a default setting repo which we can re-use. Seems like this would be useful for other projects too, and will keep them up-to-date.

https://docs.renovatebot.com/config-presets/#github-hosted-presets

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,19 +1,4 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:base",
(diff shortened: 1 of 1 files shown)


sfmig (approved):
beautiful
we are reaching automation perfection 🤖

paddyroddy:
We've exited the rabbit hole!

sfmig:
time to find a new one to dive into.... 🐰 🕳️ 🤿

paddyroddy:
Please save me

PR #186, opened 2023-10-24
Replace cookieninja with cookiecutter

p-j-smith:
- 🍪 ✂️ is being actively developed again, and  🍪 🥷 hasn't been updated since January, so let's switch back
- update docs to mention cookiecutter rather than cookieninja
- make `cookiecutter` the recommended template engine


samcunliffe (approved):
LGTM 🍉

PR #189, opened 2023-10-24
Add 3.12 to Python version lists

### LINKED ISSUE #188: Update Python version range to allow including Python 3.12
As Python 3.12 is now released we should add option to set `min_python_version` / `max_python_version` to 3.12 by adding to `cookiecutter.json` config lists.

matt-graham:
Fixes #188


p-j-smith (approved):
excellent!

PR #196, opened 2023-10-24
Use pre-commit.ci for linting repository

dstansby:
I set up https://pre-commit.ci/ to run pre-commit on the repository. The advantage of this over github actions is you can leave a comment that says "pre-commit.ci autofix" on the PR, and it will push a commit that fixes any issues that can be automatically fixed.

I left the GH actions job that lints the template after a package is generated, as I don't think there's a way to replace that.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -8,29 +8,6 @@ on:
   pull_request:
 
 jobs:
-  lint-this-repo:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout source
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
-
-      - name: Cache pre-commit
-        uses: actions/cache@v3
-        with:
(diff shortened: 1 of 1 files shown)


paddyroddy:
Personally, I'd vote for consistency 🤷

samcunliffe:
I'm 🤷 here. Do we want to recommend pre-commit.ci **_over_** running a linting job by ourselves in Actions? If so we should add that to the docs because it's a different to [what we're recommending currently](https://github.com/UCL-ARC/python-tooling/blob/main/docs/pages/ci.md).

There's also the question of (legitimately) private RSE projects which this cookie-cutter would then be less useful for.

paddyroddy:
> There's also the question of (legitimately) private RSE projects which this cookie-cutter would then be less useful for.

I would for one approach for public & private. Unfortunately, many ARC projects are still private.
CI pricing

dstansby:
👍 I buy the consistency between free/private repositories. I'll disable pre-commit.ci on this repo.

PR #197, opened 2023-10-24
Add CITATION.cff file to template

### LINKED ISSUE #22: Add CITATION.cff
Should be able to split `cookiecutter.full_name` to grab the first name and surname for CITATION.cff.
Introduced in the history of #3 but that was broken.

* Can we have multiple authors in cookie-cutter templates?
* Remove citation section from README?

matt-graham:
Fixes #22 

Adds a templated `CITATION.cff` file. Also separates the current `author_name` into separate `author_given_names` and `author_family_names` fields as Citation File Format requires these to be specified separately and it trying to split the combined name into separate fields is likely to be challenging to do in a way that covers all variants of how people write names (for example ordering of given and family names, whether additional name components like middle names are included, double-barrelled family names). I have however assumed `{given_names} {family_names}` ordering in places where we currently need the full name to be rendered - we could add a boolean option `family_name_first` or similar to allow this to be configurable.


{{cookiecutter.project_slug}}/README.md
```diff
@@ -115,24 +115,6 @@ again from the root of the repository.
...
-@article{xxx2023paper,
-  title={Title},
-  author={Author},
-  journal={arXiv},
-  year={2023}
-}
-```
-
-</details>
-
```
samcunliffe, line 135: 👍

samcunliffe (approved):
LGTM

PR #206, opened 2023-10-25
Replace `black` with `ruff-format`

### LINKED ISSUE #204: https://astral.sh/blog/the-ruff-formatter
That is all

paddyroddy:
Fixes #204


samcunliffe (changes requested):
If we're doing this, we should add `ruff-format` to the linters page as 🟢 and demote `black` to 🟠.

- [x] update pages.

paddyroddy:
> If we're doing this, we should add `ruff-format` to the linters page as 🟢 and demote `black` to 🟠.

I thought about this, but its a bit weird as its still `ruff`. Like its the same application.

samcunliffe:
> I thought about this, but its a bit weird as its still ruff. Like its the same application.

Weird slightly: but now very clear IMHO. And clarity trumps weirdness.

samcunliffe (approved):
🍉

PR #200, opened 2023-10-24
83 link each tool to the arc projects that use it or vice versa

katiebuntic:
- updating tables to link projects with each tool

Issue #54, opened 2023-06-15
Add codespell to pre-commit config


dstansby:
A suggestion to add https://github.com/codespell-project/codespell. I've found it useful in the past.

paddyroddy:
Just gave this a go. Looks good!

samcunliffe:
It is great. And we should [recommend it](https://github.com/UCL-ARC/python-tooling) for sure.

Not convinced baking it into the pre-commit is a good idea though: I've seen false positives before. And also had it break my code. (Perfectly fine, and trivial to revert... so long as it's not required in order to merge PRs or push commits.)

PR #207, opened 2023-10-25
Exclude lock files and don't just write changes

paddyroddy:
It is recommended to run the `-w` flag with also the `-i` flag https://github.com/codespell-project/codespell/blob/cb4dbe1c34911772d81b748db4c011e96ec3b54b/README.rst?plain=1#L58. I don't know how this would play with `pre-commit`, I'm tempted to remove these flags fully.

Reason being is that in #83 that this got changed https://github.com/UCL-ARC/python-tooling/pull/200/files#r1371562344 to something that it shouldn't. I've checked, and it is definitely `rouge` https://github.com/rouge-ruby/rouge.

I've also added `codespell` into the cookied repo - to keep consistent with other `pre-commit` hooks.

I've also updated the `Gemfile.lock` file - and excluded that from `codespell`.

I've also changed the YAMLs to a new way I prefer


samcunliffe:
[I think it should be dropped from pre-commit entirely.](https://github.com/UCL-ARC/python-tooling/issues/54#issuecomment-1594390696)

.pre-commit-config.yaml
```diff
@@ -6,15 +6,17 @@ repos:
...
       - id: ruff-format
-        args: ["--config", "pyproject.toml"]
+        args:
+          - --config=pyproject.toml
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
     hooks:
       - id: codespell
-        args: [--write-changes]
+        exclude: ".lock$"
```
paddyroddy, line 19: @samcunliffe can you please look at this line

samcunliffe: pre-commit will still cause the linting to fail in the case of a false positive.

.pre-commit-config.yaml
```diff
@@ -6,15 +6,17 @@ repos:
...
         exclude: "{{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}"
       - id: ruff-format
-        args: ["--config", "pyproject.toml"]
+        args:
+          - --config=pyproject.toml
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
     hooks:
       - id: codespell
-        args: [--write-changes]
```
paddyroddy, line 17: @samcunliffe they key is to just not have this line. I've used it in `UCL-MIRSG` repos recently to great effect.

.pre-commit-config.yaml
paddyroddy: Yes, but we can exclude that file in that case.

samcunliffe: So we just continually add exceptions? Maybe codespell is loads better now. It's certainly actively maintained.

I still vote that it should be dropped entirely from pre-commit and just advertised as a recommended tool.

paddyroddy: I do think you are talking a very niche case based on prior bad experience.

For the case of this repo (not the cookie generated one), I think it's only the `Gemfile.lock` that is mission-critical.

As I said I've added to the MIRSG general hooks, where we just exclude certificate files for everything https://github.com/UCL-MIRSG/.github/blob/35ec8a3074aa34903640bba9b78753ad43159b1e/precommit/mirsg-hooks.yaml#L1-L42

samcunliffe: I can handle being overruled if that's the majority decision.

dstansby (changes requested):
I think there's a few orthogonal changes, each of which deseerve their own PR so we don't get lost and to make them easier to review:

- Reformatting the `yaml`
- Removing the `--write-changes` option and excluding the lock file
- Removing `numpy` and `pydantic` from the mypy hook
- Adding `codespell` to the package template (as opposed to just using it to lint this repository)

paddyroddy:
> I think there's a few orthogonal changes, each of which deseerve their own PR so we don't get lost and to make them easier to review:

Overkill but sure

.pre-commit-config.yaml
paddyroddy: Turns out this was already being done in https://github.com/UCL-ARC/python-tooling/blob/d96dd24fbd8dfe719757b3e19dae79a36ae9086a/pyproject.toml#L1-L2 and I'm assuming the PR in question merged off `main` before that entered

paddyroddy:
Closing in favour of #210, #211, #212, #213. Have given up updating the `Gemfile.lock` as that is enough PRs as it is.

PR #210, opened 2023-10-26
Remove `codespell`

paddyroddy:
Suggestion to split #207 up

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@ repos:
     rev: v2.2.6
     hooks:
       - id: codespell
+        exclude: ".lock$"
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.4
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,6 @@ repos:
     rev: v2.2.6
     hooks:
       - id: codespell
-        exclude: ".lock$"
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.4
     hooks:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     rev: v2.2.6
     hooks:
       - id: codespell
-        args: [--write-changes]
+        exclude: ".lock$"
```
dstansby: I think it's more robust to configure this in `pyproject.toml` so it's used when `codespell` is run outside of `pre-commit`, and I did try and configure it already:
https://github.com/UCL-ARC/python-tooling/blob/d96dd24fbd8dfe719757b3e19dae79a36ae9086a/pyproject.toml#L1-L2

Is the existing config in `pyproject.toml` not working?

paddyroddy: I hadn't spotted that. I'm assuming that wasn't there at the start? I agree that is better.

If you read my original comments here https://github.com/UCL-ARC/python-tooling/pull/207#issue-1961407706 it clearly didn't work at some point. It is possible that that PR got branched off `main` and that is how the change to the lockfile happened. I don't know, and I couldn't replicate it.

dstansby (commented):
I'm 👍 to keeping write-changes, as it makes life easier and in my experience the error rate of codespell is low enough that stuff can be caught in code review. So I won't block, but I won't explicitly approve either.

samcunliffe:
In my opinion, we should also remove it from the tooling repo's pre-commit config.

paddyroddy:
Keeping @samcunliffe on board

samcunliffe (approved):

dstansby (approved):

PR #212, opened 2023-10-26
Remove `additional_dependencies`

paddyroddy:
I added these originally to make it clear how to do it for those not familiar. But perhaps it is better without.

Split up #207


dstansby (approved):
👍 makes sense, especially if a project doesn't depend on numpy or pydantic

PR #213, opened 2023-10-26
More readable YAML syntax

paddyroddy:
Suggestion to split up #207. To be merged after #210, #212.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,15 +6,16 @@ repos:
     rev: v0.1.2
     hooks:
       - id: ruff
-        args: ["--config", "pyproject.toml"]
+        args:
+          - --config=pyproject.toml
         exclude: "{{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}"
       - id: ruff-format
-        args: ["--config", "pyproject.toml"]
+        args:
+          - --config=pyproject.toml
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
     hooks:
       - id: codespell
-        args: [--write-changes]
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.4
     hooks:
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -6,15 +6,16 @@ repos:
...
         exclude: "{{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}"
       - id: ruff-format
-        args: ["--config", "pyproject.toml"]
+        args:
+          - --config=pyproject.toml
   - repo: https://github.com/codespell-project/codespell
     rev: v2.2.6
     hooks:
       - id: codespell
-        args: [--write-changes]
```
paddyroddy: See #210

{{cookiecutter.project_slug}}/.pre-commit-config.yaml
```diff
@@ -16,7 +16,6 @@ repos:
     rev: v1.6.1
     hooks:
       - id: mypy
-        additional_dependencies: [numpy, pydantic]
```
paddyroddy: See #212

samcunliffe (approved):
🤷 Meh. 

I'm fine with this, but I actually find the single-line list more readable.

dstansby (commented):
Yeah, I also find the single line (before) more readable

dstansby:
So unless there are others with a strong preference for this PR, I'd suggest we don't merge?

paddyroddy:
Think beyond the one line case.

In this example I find this a lot more readable than the alternative of `["--exclude", ".github", "--warn-list", "yaml[indentation]"` - bearing in mind that each of these can be a list. I want to use this format as I believe it's more extendable, and the `=` make it really obvious that arg2 is tied to arg1
https://github.com/UCL-MIRSG/.github/blob/de59482693728701ceb43ad8f8f9ca4b03c26bc2/precommit/mirsg-hooks.yaml#L32-L38

PR #216, opened 2023-10-30
Add `ALL` ruff rules

paddyroddy:
Not sure when `ALL` got added, but certainly better than keeping up-to-date. Have added and fixed the project accordingly.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 ci:
   autofix_prs: false
-
+exclude: "{{cookiecutter.project_slug}}"
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.1.3
@@ -8,7 +8,6 @@ repos:
       - id: ruff
         args:
           - --config=pyproject.toml
-        exclude: "{{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}"
       - id: ruff-format
         args:
           - --config=pyproject.toml
@@ -20,7 +19,6 @@ repos:
     rev: v0.23.1
     hooks:
       - id: toml-sort-fix
-        exclude: "{{cookiecutter.project_slug}}/pyproject.toml"
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.6.1
     hooks:
@@ -29,12 +27,10 @@ repos:
           - --config-file=pyproject.toml
         additional_dependencies:
           - pytest
-        exclude: "{{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}"
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v3.0.3
     hooks:
       - id: prettier
-        exclude: "{{cookiecutter.project_slug}}/.github/workflows/tests.yml"
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,8 +8,7 @@ repos:
       - id: ruff
         args:
           - --config=pyproject.toml
-        exclude: >-
-          {{ cookiecutter.project_slug }}/src/{{ cookiecutter.package_name }}
+        exclude: "{{ cookiecutter.project_slug }}/src/{{ cookiecutter.package_name }}"
       - id: ruff-format
         args:
           - --config=pyproject.toml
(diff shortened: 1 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -1,14 +1,13 @@
 ci:
   autofix_prs: false
-
+exclude: "{{ cookiecutter.project_slug }}"
```
paddyroddy, line 3: This is probably better than specifying everywhere. We have the linting of the template CI anyway.

pyproject.toml
```diff
@@ -15,51 +15,13 @@ ignore = [
     "D417", # argument description in docstring (unreliable)
     "ISC001", # simplify implicit str concatenation (ruff-format recommended)
 ]
-per-file-ignores = {"post_gen_project.py" = [
-    "T201",
-], "test_*" = [
+per-file-ignores = {"*tests*" = [
```
paddyroddy, line 18: Handling both the repo & the template

.pre-commit-config.yaml
```diff
@@ -46,7 +46,3 @@ repos:
         args:
           - --fix=lf
       - id: trailing-whitespace
-  - repo: https://github.com/renovatebot/pre-commit-hooks
-    rev: 37.36.2
-    hooks:
-      - id: renovate-config-validator
```
paddyroddy: Unnecessary for this repo and updates too frequently

pyproject.toml
```diff
@@ -15,51 +15,13 @@ ignore = [
...
-    "RUF",
-    "S",
-    "SIM",
-    "T",
-    "TCH",
-    "TID",
-    "UP",
-    "W",
-    "YTT",
+    "ALL",
```
samcunliffe, line 24: ✨

tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Test the generation of the cookiecutter template."""
```
samcunliffe: Why needed?

{{cookiecutter.project_slug}}/tests/__init__.py
```diff
@@ -0,0 +1 @@
+"""Test the {{ cookiecutter.package_name }} package."""
```
samcunliffe: Ditto.

samcunliffe (approved):
Not sure why we need `tests/__init__.py` all of a sudden(?) 

I'd do one clean PR per thing and try to keep the diff clean. So one PR replacing the list in the pyproject.toml and then one making changes to the test functions.

{{cookiecutter.project_slug}}/tests/__init__.py
paddyroddy: https://docs.astral.sh/ruff/rules/undocumented-public-package

tests/__init__.py
paddyroddy: https://docs.astral.sh/ruff/rules/undocumented-public-package

paddyroddy:
> Not sure why we need `tests/__init__.py` all of a sudden(?)

https://docs.astral.sh/ruff/rules/implicit-namespace-package

paddyroddy:
> I'd do one clean PR per thing and try to keep the diff clean. So one PR replacing the list in the pyproject.toml and then one making changes to the test functions.

Okay, but that's going to be a fair few PRs - so please do review them!

paddyroddy:
Currently, breaks [INP001](https://docs.astral.sh/ruff/rules/implicit-namespace-package). This is intended, merging.

paddyroddy:
> Okay, but that's going to be a fair few PRs - so please do review them!

Please see #219, #220 @samcunliffe

tests/__init__.py
samcunliffe: Right but our tests is not a public package.

paddyroddy: The fact that they aren't `_tests` means that they are?

paddyroddy: @matt-graham you know this stuff better than me

matt-graham: I'd agree with Sam that we ideally don't want a `__init__.py` in `tests` as my feeling is that we don't want it to be considered a package, as we'd not usually want to import test modules directly, but as `tests` will still be importable as a namespace package irrespective does make this all a bit ambiguous. [Perhaps more usefully, Pytest's suggested directory structure when having tests separate from main package doesn't include an `__init__.py`](https://docs.pytest.org/en/stable/explanation/goodpractices.html#tests-outside-application-code) so probably good to follow that? We possibly also should then follow the recommendation to add `"--import-mode=importlib"` to `tool.pytest.ini_options.addopts` in `pyproject.toml` to [avoid issues with test modules with non-unique names](https://docs.pytest.org/en/stable/explanation/goodpractices.html#which-import-mode).

matt-graham: I missed that this is now closed and conversation is going on elsewhere so ignore above.

samcunliffe: My bad. Poor thread discipline, and poor redirection.
→ https://github.com/UCL-ARC/python-tooling/pull/219#issuecomment-1794687931 ff

paddyroddy: I'm going to add an issue for @matt-graham's suggestion!

PR #219, opened 2023-11-06
Fix `INP001`

paddyroddy:
https://docs.astral.sh/ruff/rules/implicit-namespace-package

Continuation of #216 see [comment](https://github.com/UCL-ARC/python-tooling/pull/216#issuecomment-1794555506)

Note this will fail due to [D104](https://docs.astral.sh/ruff/rules/implicit-namespace-package)


dstansby:
I'm 👎 for adding extra files that aren't needed just to fix a ruff rule

samcunliffe:
Yeah, right. I don't understand why we need these. Our tests are not a public package, we don't ever do `from package.tests import flibble`, and `pytest` finds both the repo and the cookie-cut tests.

~~Adam once explained a genuine need for `tests/__init__.py` but I now can't find where that was.~~
[Sometimes we need it](https://github.com/SainsburyWellcomeCentre/WAZP/pull/36#discussion_r1137632426), but [`pytest` still say to omit](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery).

I'd probably err on the side of adding an exception to INP001 for the tests dir. But I don't have an incredibly strong opinion.

paddyroddy:
Honestly, I don't care. However, my understanding is that `tests` is a public package because it isn't called `_tests`.

paddyroddy:
Found astral-sh/ruff#6474, so have removed `__init__.py` from `tests`. How do we feel about `hooks`? Also, in different.

samcunliffe:
> How do we feel about `hooks`? Also, in different.

Very very very soft opinion, is to add it as another exception. It seems [cookiecutter hooks](https://cookiecutter.readthedocs.io/en/stable/advanced/hooks.html) (like git hooks) are supposed to be closer to a directory of executables. So... not a package. I mean: we _could_ put bash scripts or whatever in there. (Obviously we won't do that, we're not crazy.)

But I'm also OK with leaving as-is. It's only _this_ repo code. We're not adding bloat to users' packages.

paddyroddy:
> Very very very soft opinion, is to add it as another exception. It seems [cookiecutter hooks](https://cookiecutter.readthedocs.io/en/stable/advanced/hooks.html) (like git hooks) are supposed to be closer to a directory of executables. So... not a package. I mean: we _could_ put bash scripts or whatever in there. (Obviously we won't do that, we're not crazy.)

Having talked to @dpshelio, I'm inclined to remove the `hooks/__init__.py`

paddyroddy:
Can you review @samcunliffe?

samcunliffe (approved):

PR #220, opened 2023-11-06
Fix `D104`

paddyroddy:
https://docs.astral.sh/ruff/rules/undocumented-public-package

Continuation of #216 see [comment](https://github.com/UCL-ARC/python-tooling/pull/216#issuecomment-1794555506)

To be merged after #219

### DIFF
--- a/hooks/__init__.py
+++ b/hooks/__init__.py
@@ -1,0 +1,1 @@
+"""Scripts executed at specific stages during the project generation process."""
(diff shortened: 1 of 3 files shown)


samcunliffe:
Pause pending discussion in #219.

I think we _do_ want to keep D104 in general though. `\{\{cookiecutter.package_name\}\}/__init__.py` should be documented.

paddyroddy:
Not required any more, closing

PR #247, opened 2024-01-23
Create CODE_OF_CONDUCT.md

paddyroddy:
Have used a template. Do we want one for the template too?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -17,23 +17,23 @@ diverse, inclusive, and healthy community.
 Examples of behavior that contributes to a positive environment for our
 community include:
 
-* Demonstrating empathy and kindness toward other people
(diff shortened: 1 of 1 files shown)


samcunliffe:
I think we _should_ use the code of conduct here. But not force it on template users. Which is as you've proposed as of c9c5a44.

Perhaps we can go a bit rogue and add a link to the [UCL codes of conduct](https://www.ucl.ac.uk/research-ethics/guidance-training-and-resources/guidance/codes-conduct).

paddyroddy:
> Perhaps we can go a bit rogue and add a link to the [UCL codes of conduct](https://www.ucl.ac.uk/research-ethics/guidance-training-and-resources/guidance/codes-conduct).

I don't mind, as long as it seems relevant to coding

samcunliffe (approved):
🫐

Issue #246, opened 2024-01-23
Advertise the repo.


samcunliffe:
- [x] Create a social banner (a la [matt-graham/mici](https://github.com/matt-graham/mici)
- [x] Finalise logo.
- [x] #330 
----

- [x] @matt-graham tutorial at "The Festival of Digital Research and Scholarship".
- [x] @Saransh-cpp 's `uv`ified tutorial at the "Second Festival of Digital Research and Scholarship".
- [x] CIO family demo.
- [ ] Post on RSE Slack.
- [ ] Post on Research Programming Hub.
- [ ] CIO family teams.
- [ ] UCL collaborations hour (speaker?)
- [ ] Make poster

samcunliffe:
I'm going to close this since it's somewhat completed. We can always refer back to our list of grand plans if we ever want to do another of them.

PR #277, opened 2024-01-23
Add logo and banner

matt-graham:
Following on from https://github.com/UCL-ARC/python-tooling/discussions/144, adds SVG files for logo we decided on in meeting today to a new `images` subdirectory plus a corresponding social banner to address task in #246. The text colour on the banner should be visible on both dark and light backgrounds.

I've also configured the Jekyll site to show the logo in the header bar - by default this was very small so I've added some custom (S)CSS to adjust the header div size and center the logo.

Currently only the logo is shown in the header not any text - not sure if it is possible to override this?


docs/_config.yml
```diff
@@ -6,3 +6,5 @@ url: https://UCL-ARC.github.io/python-tooling
 
 aux_links:
   Source repository: https://github.com/UCL-ARC/python-tooling
+
+logo: "https://raw.githubusercontent.com/UCL-ARC/python-tooling/main/images/logo.png"
```
matt-graham: I couldn't get a relative path to work locally so used full URL - not sure if we could put `images` somewhere under `docs` directory to get this working and/or if there are relevant Jekyll / theme configuration options.

docs/_sass/custom/custom.scss
```diff
@@ -0,0 +1,8 @@
+div.site-header {
+  max-height: 15em;
+  height: 15em;
```
matt-graham, line 3: As `max-height` attribute is set in theme styling found I needed to override both `max-height` and `height` for this to work but someone with more CSS knowledge may know a better approach.

images/logo.png
matt-graham, line 1: We could potentially remove this. I initially thought it might be necessary to have a raster image for including in theme but rendering SVG directly works. It can be useful to have a PNG for some use cases but easy enough to just generate from SVG.

images/banner.svg
samcunliffe, line 1: Should it go at the top of the README? We can replace the title with the banner, right?

docs/_config.yml
samcunliffe: Possibly the images are not included in the artifact upload step (therefore not included in the website deployment??).

I'd say this way is probably fine.

images/banner.svg
matt-graham: Having the logo in the README seems a good idea though from a digital accessibility standpoint might be better to have title still as text (though I guess `alt` text would somewhat solve that). Could have logo above header / title - something like https://github.com/matt-graham/symnum?

docs/_config.yml
paddyroddy: I think you have to do this kind of thing for relative links https://github.com/paddyroddy/paddyroddy.github.io/blob/347e094cb0af0c609394b2beaa90770929feca1e/running.html#L14.

I think your approach is fine, though.

images/banner.svg
paddyroddy: > Could have logo above header / title - something like [matt-graham/symnum](https://github.com/matt-graham/symnum)?

That looks nice

docs/_config.yml
paddyroddy: If keeping this, quotes add nothing
```suggestion
logo: https://raw.githubusercontent.com/UCL-ARC/python-tooling/main/images/logo.png
```

paddyroddy (approved):
Yay 🫐

docs/_config.yml
matt-graham: My understanding was that its [recommended to quote strings including special characters like `:`](https://stackoverflow.com/a/22235064) in YAML files and the `just-the-docs` [documentation example included quotes ](https://just-the-docs.github.io/just-the-docs/docs/configuration/#site-logo) so I followed this.

paddyroddy: IMO that seems very cautious. I've quite enjoyed removing them, and only using them when needed. So for templating

Needed:
```
variable_a: "{{some_var}}/path/to/folder"
```

Not needed:
```
variable_b: path/to/folder/{{some_var}}
```

I find it easier to read without, so only include when actually required. But I'm indifferent as to what you do.

matt-graham: Thanks useful to know for future reference!

images/banner.svg
matt-graham: I've had a go at replicating this in the README - unfortunately to get image to be centre aligned need to fallback to HTML rather than Markdown. I've also centred the heading at the moment but could keep this left aligned and just centre logo. As the image URL includes the branch and the logo is not currently on `main` (and I guess we don't want to keep this branch around) this won't currently render correctly until merged.

matt-graham: Oh and the image not existing yet is causing the link checker to fail. I guess probably worth separating out in a different PR - will revert here.

samcunliffe:
🫐

matt-graham:
I made an aborted attempt to include a favicon version of logo as noticed Jekyll complained about not being able to find `favicon.ico` when serving but this didn't seem to work when serving locally so have removed. Also in the course of testing locally I ended up switching the branch in the URL to `mmg/add-logo-and-banner` but have now reverted that back to main. Given it looks like including in the README here will be a bit 🐔 / 🥚 due to the link checker / image URL not existing yet, I think we can do in a separate PR. @paddyroddy / @samcunliffe just to confirm given I've made several changes since you reviewed, are we okay to merge this?

paddyroddy:
I'm happy, merging

Issue #273, opened 2024-01-23
[TASK]: Test if the issue templates work in the template


paddyroddy:
### What needs to be done?

I need to check when you make a new project what happens

paddyroddy:
They do!

PR #286, opened 2024-01-24
Fix issue template bug report number

paddyroddy:
Have verified that #273 works. But the bug report wasn't quite right, this fixes that.

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/{{cookiecutter.project_slug}}/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -52,8 +52,7 @@ body:
       description: |-
         Can be found by
         ```sh
-        python -c "import {{cookiecutter.package_name}};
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -52,8 +52,7 @@ body:
       description: |-
         Can be found by
         ```sh
-        python -c "import {{cookiecutter.package_name}};
-                   print({{cookiecutter.package_name}}.__version__)"
+        python -c "import {{cookiecutter.package_name}}; print({{cookiecutter.package_name}}.__version__)"
```
paddyroddy, line 55: This was me loving wrapping, but then copy-and-paste wouldn't work 😭

samcunliffe: 🎁

samcunliffe (approved):

PR #287, opened 2024-01-24
Fix linting the template check

paddyroddy:
Noticed during my test of issue templates #273, that this was actually not working!

This should fail, due to dodgy linting.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -55,5 +55,7 @@ jobs:
         run: cookiecutter . --no-input --output-dir cookie-template
 
       - name: Run pre-commit
-        run: pre-commit run --all-files --color always --verbose
-        working-directory: cookie-template
+        run: |-
+          git add .
+          pre-commit run --all-files --color always --verbose
+        working-directory: cookie-template/python-template
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
           - pytest
         exclude: |-
           (?x)(
-            {{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}
+            {{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}|
             {{cookiecutter.project_slug}}/tests
           )
   - repo: https://github.com/pre-commit/mirrors-prettier
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,8 @@ repos:
           - pytest
         exclude: |-
           (?x)(
-            {{cookiecutter.project_slug}}/(src/{{cookiecutter.package_name}}|tests)
+            {{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}
+            {{cookiecutter.project_slug}}/tests
           )
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v3.1.0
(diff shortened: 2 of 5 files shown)


paddyroddy:
Yay it failed!

{{cookiecutter.project_slug}}/LICENSE.md
```diff
@@ -649,4 +649,4 @@ what you want to do, use the GNU Lesser General Public License instead of this
 License. But first, please read
 &lt;<http://www.gnu.org/philosophy/why-not-lgpl.html>&gt;.
 
-{%- endif -%}
+{%- endif %}
```
paddyroddy, line 652: Thanks to @dpshelio for helping me on this one!

.pre-commit-config.yaml
```diff
@@ -5,7 +5,10 @@ repos:
       - id: ruff
         args:
           - --config=pyproject.toml
-        exclude: "{{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}"
+        exclude: |-
+          (?x)(
```
paddyroddy, line 9: I've changed to this syntax in the case that we need to ignore more than one thing

.github/workflows/linting.yml
```diff
@@ -55,5 +55,7 @@ jobs:
         run: cookiecutter . --no-input --output-dir cookie-template
 
       - name: Run pre-commit
-        run: pre-commit run --all-files --color always --verbose
-        working-directory: cookie-template
+        run: |-
+          git add .
```
samcunliffe, line 59: 🐛-fix here  ☝️ ?

.pre-commit-config.yaml
samcunliffe: I don't get it. Why are we excluding this all of a sudden? And why all over the shop?

samcunliffe (commented):
I'm missing something...

.github/workflows/linting.yml
paddyroddy: That and the `working-directory`. We were never actually linting the template before.

.pre-commit-config.yaml
paddyroddy: If you look, it's the exact same `exclude`s that we had before. It's just the format which has changed.

Unfortunately, @pre-commit isn't using an array for `exclude`, so you have to rely on this ugly expression https://pre-commit.com/#regular-expressions

paddyroddy:
Merged in `main`, which I knew would break it. It was in fixing #286 that I realised linting wasn't working. Fixing now.

paddyroddy:
Fixed now

.pre-commit-config.yaml
```diff
@@ -26,12 +32,18 @@ repos:
           - --config-file=pyproject.toml
         additional_dependencies:
           - pytest
-        exclude: "{{cookiecutter.project_slug}}/(src/{{cookiecutter.package_name}}|tests)"
+        exclude: |-
+          (?x)(
+            {{cookiecutter.project_slug}}/(src/{{cookiecutter.package_name}}|tests)
```
dstansby:
```suggestion
            {{cookiecutter.project_slug}}/src/{{cookiecutter.package_name}}
            {{cookiecutter.project_slug}}/tests
```
Since there's the nice new regex sytax, I think it's a lot clearer splitting this on to two lines?

paddyroddy: Yeah, fair enough, I had wondered about this

stefpiatek (approved):

paddyroddy:
Someone reviewed it 🤯 thanks @stefpiatek!

PR #289, opened 2024-01-24
Consistency in LICENSE file extension in template

paddyroddy:
Currently, we have `LICENSE` at the top level, and `LICENSE.md` in the template.


{{cookiecutter.project_slug}}/LICENSE
samcunliffe, line 1: Ah no. If we do this, we aught to go the other way. The LICENSE.md inside the cookie-cutter template is actually markdownified.

Easiest is probably to make the top-level repo LICENSE be markdown.

samcunliffe (changes requested):

{{cookiecutter.project_slug}}/LICENSE
paddyroddy: Ah, sorry, hadn't spotted/realised that. I wasn't sure how much you could modify a licence, so recently changed my stuff to `.txt` (as not a fan of no extension). I'll raise an issue to do this.

paddyroddy: #291

Issue #291, opened 2024-01-24
[TASK]: Markdownify the top level licence


paddyroddy:
### What needs to be done?

For consistency, it would be nice to have the top level `LICENSE` file, be a `LICENSE.md` file in the same way as the template. Essentially do the opposite of #289.

PR #293, opened 2024-01-25
Top-level LICENSE → LICENSE.md

samcunliffe:
Title changed to be consistent with [SPDX](https://spdx.org/licenses/MIT.html) and [choosealicense](https://choosealicense.com/licenses/mit/).

NAL but I think the title isn't part of the license and the exact phrasing "doesn't matter" (but let's have it consistent and correct anyway).

Solves 
* #291

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-# The MIT License (MIT)
+# MIT License
 
 Copyright (c) 2023-2024 UCL Centre for Advanced Research Computing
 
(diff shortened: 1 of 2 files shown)


dstansby (approved):

paddyroddy:
> NAL

What does this mean❓

samcunliffe:
Soz. Full form: [IANAL](https://en.wikipedia.org/wiki/IANAL).

paddyroddy:
> Soz. Full form: [IANAL](https://en.wikipedia.org/wiki/IANAL).

Thank you. Would never have worked that one out.

PR #297, opened 2024-02-06
Remove default prefixes from issue titles

dstansby:
Personally I find the list of issues much harder to read when it has these all upper case prefixes, and since issues are labelled with their category anyway, adding these prefixes is duplicating that info.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 
 #default project path
 python-template
+
+*.DS_Store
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -5,3 +5,5 @@
 
 #default project path
 python-template
+
+*.DS_Store
```
samcunliffe, line 9: Thanks.

samcunliffe (approved):
I agree with you. Labels are already there and are better.

ALSO THE LABELS AREN'T SHOUTING.

PR #302, opened 2024-02-06
Add a [test] extra to package dependencies

dstansby:
This

- prevents dependencies being duplicated in the package requirements and the `tox` config.
- makes it easier to install `tox` (in the dev extra) without installing pytest, which isn't needed if you're just going to run tests via tox.

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -41,7 +41,6 @@ test = [
     "pytest-cov",
 ]
 dev = [
-    "{{cookiecutter.project_slug}}[test]",
     "build",
     "mypy",
     "pre-commit",
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -30,9 +30,6 @@ dynamic = [
 keywords = [
 ]
 name = "{{cookiecutter.project_slug}}"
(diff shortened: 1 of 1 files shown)


dstansby:
Anyone know how I can lint the template locally? It's not clear to me what needs fixing from the failed GH actions build.

paddyroddy:
> Anyone know how I can lint the template locally? It's not clear to me what needs fixing from the failed GH actions build.

Unfortunately, the only real way is by making a dummy project. Something like this as it's simplest https://github.com/UCL-ARC/python-tooling/blob/32786c9a29ed379303bd3298396071cc9176ec49/.github/workflows/linting.yml#L55-L60

paddyroddy:
At a glance, I can tell it will be `toml-sort`. Can fix for you now

paddyroddy (approved):
🫐

dstansby:
Was moving this to `[project.optional-dependencies]` a technical issue, or do you prefer it all to be under `[project]`?

dstansby:
Thanks for fixing!

paddyroddy:
> Was moving this to `[project.optional-dependencies]` a technical issue, or do you prefer it all to be under `[project]`?

Just preference. I like the idea that each section is clearly about one thing, i.e. `ruff`, `pytest`, `project`, etc.

paddyroddy:
The "problem" was just the order of `test` and `dev` 😛

PR #303, opened 2024-02-14
Add `lint` prefix to `ruff` configuration

paddyroddy:
This will be encouraged/forced in `v0.2.0`

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -72,7 +72,7 @@ testpaths = [
 [tool.ruff]
 fix = true
 force-exclude = true
-ignore = [
+lint.ignore = [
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -72,27 +72,26 @@ testpaths = [
...
-per-file-ignores = {"tests*" = [
+lint.per-file-ignores = {"tests*" = [
     "INP001",
     "S101",
 ]}
-select = [
+lint.select = [
     "ALL",
 ]
-target-version = "py{{cookiecutter.min_python_version | replace('.', '')}}"
```
paddyroddy, line 90: [`requires-python` is recommended over `target-version`](https://docs.astral.sh/ruff/settings/#target-version), and we don't need both.

dstansby (approved):

PR #306, opened 2024-02-16
Add a placeholder title to all issue templates.

samcunliffe:
I was today years old when I learned this:

Screenshot 2024-02-16 at 07 45 54

---

Fixes [this problem](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#key-must-be-a-string) introduced in 
- #297

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,7 @@
 ---
 name: Bug Report
 description: Create a Report to Help us Improve
-title: ""
+title: "A one-line description of your problem"
 labels:
   - bug
 assignees:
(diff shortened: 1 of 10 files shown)


paddyroddy (changes requested):
This is new, definitely worked before but can see it's not working now. Can you add to the template too?

paddyroddy:
> This is new, definitely worked before but can see it's not working now. Can you add to the template too?

Actually, this is wrong. I had `"[BUG]: "` before, it's the act of removing those that broke stuff

samcunliffe:
> This is new, definitely worked before but can see it's not working now. Can you add to the template too?

Done.

paddyroddy (approved):

PR #307, opened 2024-02-16
Validate GitHub issue templates

dstansby:
This validates the GitHub issue tempaltes against a schema, and would have caught the issue that was fixed by https://github.com/UCL-ARC/python-tooling/pull/306. 

Unfortunately it's not possible to validate the package template files as they contain invalid characters for a string entry (`{` and `}`).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,3 +56,12 @@ repos:
         args:
           - --fix=lf
       - id: trailing-whitespace
+  - repo: https://github.com/python-jsonschema/check-jsonschema
+    rev: cb934cfd6d2eea1aed7b20e4e9e6339d5952322c # 0.28.0
+    hooks:
+      # Schemas taken from https://www.schemastore.org/json/
+      - id: check-jsonschema
+        name: "Validate GitHub issue templates"
+        files: ^\.github/ISSUE_TEMPLATE/.*\.yml$
+        exclude: ^\.github/ISSUE_TEMPLATE/config\.yml$
+        args: ["--verbose", "--schemafile", "schemas/github-issue-forms.json"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,7 +57,7 @@ repos:
           - --fix=lf
       - id: trailing-whitespace
   - repo: https://github.com/python-jsonschema/check-jsonschema
-    rev: cb934cfd6d2eea1aed7b20e4e9e6339d5952322c # 0.28.0
+    rev: 0.28.0
     hooks:
       # Schemas taken from https://www.schemastore.org/json/
       - id: check-jsonschema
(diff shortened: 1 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -56,3 +56,12 @@ repos:
...
       - id: trailing-whitespace
+  - repo: https://github.com/python-jsonschema/check-jsonschema
+    rev: cb934cfd6d2eea1aed7b20e4e9e6339d5952322c # 0.28.0
+    hooks:
+      # Schemas taken from https://www.schemastore.org/json/
+      - id: check-jsonschema
+        name: "Validate GitHub issue templates"
+        files: ^\.github/ISSUE_TEMPLATE/.*\.yml$
+        exclude: ^\.github/ISSUE_TEMPLATE/config\.yml$
+        args: ["--verbose", "--schemafile", "schemas/github-issue-forms.json"]
```
paddyroddy: Why can you add this to the `.pre-commit-config.yaml` of the template? Then they both get validation

schemas/github-issue-forms.json
```diff
@@ -0,0 +1,2377 @@
+{
```
paddyroddy, line 1: Is it necessary to add this to the repo? Like is there a way to do this remotely?

paddyroddy (changes requested):

.pre-commit-config.yaml
dstansby: > Unfortunately it's not possible to validate the package template files as they contain invalid characters for a string entry ({ and }).

schemas/github-issue-forms.json
dstansby: I don't think there's another way

.pre-commit-config.yaml
dstansby: Ah sorry, just realised what you meant! Good shout, I will do that.

dstansby: 👍 done

paddyroddy: I did wonder if you'd missed something

.pre-commit-config.yaml
```diff
@@ -56,3 +56,12 @@ repos:
         args:
           - --fix=lf
       - id: trailing-whitespace
+  - repo: https://github.com/python-jsonschema/check-jsonschema
+    rev: cb934cfd6d2eea1aed7b20e4e9e6339d5952322c # 0.28.0
```
paddyroddy: I assume `renovate` will use this syntax in the future. How does this play with `pre-commit autoupdate`?

paddyroddy (commented):
Sorry thought of one more thing

.pre-commit-config.yaml
dstansby: 😢 `pre-commit autoupdate` doesn't like it, and just changes it back to a version number. So I've switched the SHAs to version numbers.

paddyroddy: Yeah, I thought that might be the case. The `pre-commit` maintainer is pretty stubborn, too, so he likely won't change the format.

paddyroddy (approved):

PR #308, opened 2024-02-16
Don't add an asignee in the package issue template

### LINKED ISSUE #305: `github_user` does not indicate value can be an organisation
I didn't realise that the intent of the `github_user` variable was to be the username or organisation where the git repository is remotely located. So when I ran the cookiecutter I put in my username, then had to change it all to the org I was actually using.

I think this is best solved by solving https://github.com/UCL-ARC/python-tooling/issues/296.

dstansby:
The motivation behind this is:

1. In my experience it's rare for a single person to be assigned every issue in a repository. Much more common are no-one is assigned by default, multiple people (e.g. core developers) are assigned by default, or a group is assigned by default.
2. Not having a GitHub user in the issue template makes it much easier to fix https://github.com/UCL-ARC/python-tooling/issues/305 by changing the `github_user` field to stand for either a user *or* organisation.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -59,4 +59,4 @@ body:
         - Cookiecutter version:
         - Operating system:
         - Python version:
-      render: markdown
+      render: Markdown
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,3 +56,12 @@ repos:
         args:
           - --fix=lf
       - id: trailing-whitespace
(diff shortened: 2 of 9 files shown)


paddyroddy (changes requested):
Are we sure it will work when `assignees` isn't present altogether? Should we be doing
```yaml
assignees:
```

dstansby:
I've rebase on to the schema validation PR, and removing assignee altogether passes the validation.

dstansby:
Whoever merges please squash merge!

paddyroddy (approved):

PR #309, opened 2024-02-19
Correct uses of project_name in pyproject.toml

dstansby:
1. Fix `[tool.coverage]` so coverage data is collected correctly. I haven't come up with a way to test this, other than this is the fix I needed at https://github.com/hipCTProject/hipct-reg
2. While I was at it, I noticed a second place where custom logic could just be repalced with project_name


paddyroddy (approved):
Good spot

PR #311, opened 2024-02-21
Don't try and measure coverage of .tox installation

dstansby:
I'm not too sure why this line is here, but I don't *think* it's needed since either:
- You're running `pytest` and the source code is in ./src
- You're running `tox` and tox installs the package in it's own virtual environment, and the source code is still at ./src relative to the installed pyproject.toml file

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -53,7 +53,6 @@ run = {branch = true, parallel = true, source = [
 ]}
 paths.source = [
     "src",
-    ".tox*/*/lib/python*/site-packages",
 ]
 
 [tool.mypy]


paddyroddy (changes requested):
This is needed (in my experience, battled on this for hours at some point) for the `pytest-cov` output to have the paths as `src/...` rather than `.tox/...../src/..`

dstansby:
Can you share the package where you needed this? As far as I know nothing should be referencing files in `.tox` as they are temporary and manged by tox itself.

paddyroddy:
> Can you share the package where you needed this? As far as I know nothing should be referencing files in `.tox` as they are temporary and manged by tox itself.

It's going to take a while to find, this was over a year ago

paddyroddy:
There was a blog where I fixed my problem.

[tox themselves do this](https://github.com/tox-dev/tox/blob/c024fe243b42d8db4ad52e3a8faaa63649319356/pyproject.toml#L154-L161).

dstansby:
👍 just been down this rabbit hole, and it seems safer not to touch this if it works (given I don't think there's any tests for coverage being correct...)

paddyroddy:
If I find the blog I'll put it here

paddyroddy:
Not sure if this was the one, but it explains the problem https://hynek.me/articles/testing-packaging

paddyroddy:
Yeah that was the blog, found it pretty easily in the end

dstansby:
FWIW I solved my issue by setting `usedevelop = true` in `tox.ini`. This made sure the test files and package source that was being imported was in-place, and not from the tox environment packages directory.

Thanks for the link!

paddyroddy:
What actually was your issue?

dstansby:
I was getting coverage on library files, but not on test files. It seems like tox was picking up the test files from outside the installation in `.tox` for some reason... `usedevelop=true` fixes this by leaving all the source (tests and library code) in the source folder, and just doing an editable install into the tox environment. That way just files under `src` can be checked to get complete coverage.

paddyroddy:
Okay, think I get you

PR #312, opened 2024-02-21
Don't skip reporting files with full code coverage

dstansby:
It seems sensible to report all files, even if they have 100% test coverage.


paddyroddy:
There's no harm. But IMO I like to see easily what needs coverage adding. Removes needless clutter.

samcunliffe:
Hmm. I see your point, but for relative newcomers and people who don't want to bother about configuring things (who are kinda our target audience), the subtlety of this setting might be confusing.

samcunliffe (approved):

PR #314, opened 2024-03-07
Dark mode.

samcunliffe:
Changes the docs pages to dark mode by default and adds a switcher button to the top bar.

The first time around there is an explicit "View these pages in light mode" then subsequently, it's the sun and moon characters.

https://github.com/UCL-ARC/python-tooling/assets/1836192/b003bfd8-fa35-444f-8e41-0ee77173363d

## Up for debate

My proposal is this. But I made two decisions (which you might disagree with).

* Dark mode by default.
* Words first, then ☼/☾ characters.

JTD doesn't seem to support just taking the system theme (yet).
- https://github.com/just-the-docs/just-the-docs/issues/1223

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -4,6 +4,8 @@ theme: just-the-docs
 
 url: https://UCL-ARC.github.io/python-tooling
 
+color_scheme: dark
+
 aux_links:
   Source repository: https://github.com/UCL-ARC/python-tooling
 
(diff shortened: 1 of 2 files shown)


dstansby (approved):
I'm more pro just having the icons and no words, but not by much. Will approve but not merge to give others time to have opinions on this.

paddyroddy (approved):
I like it. But yeah I don't like the words - however, if someone isn't used to the 🌙 🌞 convention, I can see an argument for it

samcunliffe:
Three votes to remove the words. 🗳️👨‍⚖️

PR #316, opened 2024-03-18
Remove mentions of mamba

### LINKED ISSUE #301: Remove mentions of mamba?
[libmamba is now the default solver for conda](https://conda.org/blog/2023-11-06-conda-23-10-0-release/), so to keep things simple it might be worth removing mentions of mamba from the docs here?

dstansby:
Since the default solver in conda is now libmamba, remove mentions of mamba to simplify recommendations.

I also cleaned the conda links so they point to stable docs (not latest), and did "conda" > "Conda" where appropriate.

Fixes https://github.com/UCL-ARC/python-tooling/issues/301

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,8 @@ Python packages with our recommended tooling set up and ready to go.
 
 ## Using this template
 
-1.  Install [cookiecutter] in a `conda` or `venv` environment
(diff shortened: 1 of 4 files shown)


paddyroddy (approved):
🐍

PR #318, opened 2024-03-20
Adding Sphinx docs to template

### LINKED ISSUE #16: Add documentation
A really lazy way: paddyroddy/python-template#9 (+ paddyroddy/python-template#10 + paddyroddy/python-template#11) as I'm an idiot) which allows you to just use `github-pages` deploying from a `gh-pages` branch with these settings
image

A better example that I just did looks like: https://astro-informatics.github.io/sleplet/

matt-graham:
Would resolve #16 

Adds Sphinx documentation to template using `sphinx-autodoc2` to generate API documentation and using MyST parser to allow use of Markdown in docstrings (and documentation pages). Also sets up a few other extensions - [`sphinx.ext.intersphinx`](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) for automatic links to other documentation with by default only the Python standard library index included, and [`sphinx.ext.viewcode`](https://www.sphinx-doc.org/en/master/usage/extensions/viewcode.html) to add links to source code. This is set up to use the [PyData Sphinx theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/) which @samcunliffe recommeded on another project. A `tox` environment `docs` is added to simplify building documentation. 

This was mainly intended as a demonstration that it's possible to get Sphinx set up in a way which will easy for users to use and doesn't require much boilerplate for reference in discussion in #16. It's not clear yet if we would want to go with Sphinx rather than Mkdocs.

I also haven't yet added any default workflows to the template for building docs - ideally we'd want to make it easy to automatically deploy to GitHub pages but this would require some manual intervention from user to specify where GitHub pages site should be generated from. We could as an intermediate step just have a workflow for testing docs build, with a commented out step for subsequently doing deployment on pushes to `main` with instructions to user to do necessary configuration before uncommenting.

EDIT: Screenshot of what generated documentation website looks like

image

To dos
- [x] Add workflow for testing building documentation on pull-requests and pushes to main (for example https://github.com/UCL/neso-calibration/blob/main/.github/workflows/docs.yml)
- [ ] Add instructions of how to set up automatic deployment to GitHub pages in main README or as option in cookiecutter similar to `git init`
- [x] Add instructions for how to build documentation to template README.md
- [ ] Decide whether to include README.md content in docs

### DIFF
--- a/{{cookiecutter.project_slug}}/.github/workflows/docs.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/docs.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .tox
-          key: tox-${{ '{{' }} hashFiles('pyproject.toml')${{ '}}' }}
+          key: tox-${{ '{{' }} hashFiles('pyproject.toml') {{ '}}' }}
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
@@ -29,13 +29,13 @@ jobs:
         run: tox -e docs
     # Uncomment lines below to set-up automatic deployment of documentation to a
     # GitHub Pages website on pushing to main - you will need configure the repository
-    # to deploy from a branch `gh-pages` (https://tinyurl.com/gh-pages-from-branch),
+    # to deploy from a branch gh-pages (https://tinyurl.com/gh-pages-from-branch),
     # which will be created the first time this workflow step is run
     #   - name: Publish documentation on GitHub pages
     #     if: success() && github.event_name != 'pull_request'
     #     uses: peaceiris/actions-gh-pages@v3
     #     with:
-    #       github_token: ${{ '{{' }}  secrets.GITHUB_TOKEN ${{ '}}' }}
+    #       github_token: ${{ '{{' }}  secrets.GITHUB_TOKEN {{ '}}' }}
     #       publish_dir: docs/_build/html
     #       publish_branch: gh-pages
     #       user_name: "github-actions[bot]"
--- a/{{cookiecutter.project_slug}}/.github/workflows/docs.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/docs.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .tox
-          key: tox-${{hashFiles('pyproject.toml')}}
+          key: tox-${{ '{{' }} hashFiles('pyproject.toml')${{ '}}' }}
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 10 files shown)


samcunliffe:
>  Decide whether to include README.md content in docs

I do quite like this. Sphinx vs MkDocs aside. It's DRY.

{{cookiecutter.project_slug}}/docs/_static/.gitignore
dstansby, line 1: I think if you got rid of the line `html_static_path = ["_static"]` from `conf.py` you could get rid of this file?

matt-graham: Thanks - for some reason I thought we had to set `html_static_path` to something but have removed as you suggested and all works fine.

paddyroddy: The very fact that someone as experienced you isn't sure - highlights the difficulties of `sphinx`

{{cookiecutter.project_slug}}/.github/workflows/docs.yml
```diff
@@ -0,0 +1,42 @@
...
+    # which will be created the first time this workflow step is run
+    #   - name: Publish documentation on GitHub pages
+    #     if: success() && github.event_name != 'pull_request'
+    #     uses: peaceiris/actions-gh-pages@v3
+    #     with:
+    #       github_token: ${{ '{{' }}  secrets.GITHUB_TOKEN ${{ '}}' }}
+    #       publish_dir: docs/_build/html
+    #       publish_branch: gh-pages
+    #       user_name: "github-actions[bot]"
+    #       user_email: "github-actions[bot]@users.noreply.github.com"
```
samcunliffe: Do we need a step that does something like: `touch .nojekyll` to [bypass jekyll](https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/)? Or is that handled somewhere?

Could also commit a file.. but I don' think we need to. Could all be done on the CI, right?
https://github.com/UCL/dxh/commit/863ce7a3ba7176f0e6f6567244a46270dd8b24fe

samcunliffe: Again, this is build-tool-agnostic.

matt-graham: The `peaceiris/actions-gh-pages@v3` action handles adding a `.nojekyll` file automatically (unless you pass an `enable_jekyll: true` option).

paddyroddy: This is what I've also typically done @samcunliffe https://github.com/paddyroddy/.github/blob/c3bb562d91f4004231cdd8497106a209a7531f40/actions/github-pages/jekyll/action.yml#L44. However, we should keep it consistent in this project, whatever we do.

matt-graham:
Closing in favour of #319 as consensus decision in #16 is to add MkDocs instead to template.

PR #319, opened 2024-03-20
Adding MkDocs docs to template

matt-graham:
Would resolve #16 as an alternative to #318

Adds MkDocs documentation to template, aiming for a comparable set of features to what is implemented in #318 - automatically built API documentation with Markdown syntax, support for references to external documentation inventories, dark / light mode toggle, GitHub repository link in docs, workflow for automatically testing building docs.

Example screenshot of rendered index page

image

and of API reference page

image

PR #320, opened 2024-03-20
Always ignore template at top level pre-commit

paddyroddy:
We lint at the top level, and at the template level. It doesn't really make sense to template both, and some of the hooks get confused with the `cookiecutter` templating. Previously, we were ignoring some, but actually makes sense to ignore it all.

The downside is that locally it will be annoying to lint at the package level. But that is fine because CI with lint at the template level. Sometimes you may need to create a dummy project to fix the issue.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+exclude: ^{{cookiecutter.project_slug}}/
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.1.15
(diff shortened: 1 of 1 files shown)


dstansby (commented):
So I understand, on one of the CI runs we still template a new package and run pre-commit on that new package? If so I think this is a nice simplification 👍

matt-graham (approved):
From experience in working on both #318 and #319 the occasional bad interactions between templating syntax and what pre-commit hooks expect in the various files are a bit of pain so this seems a good idea to me.

samcunliffe (approved):
Yeah, agreed.

PR #323, opened 2024-03-23
Add a page on parallel and async tools.

### LINKED ISSUE #178: Add a page on tools for parallel programming

samcunliffe:
Solves #178.

No async yet, but there are a couple of projects using async so I'll ask around for opinions.

### DIFF
--- a/docs/pages/parallel-async.md
+++ b/docs/pages/parallel-async.md
@@ -23,7 +23,7 @@ simply due to pre-existing code using a library like [pandas].
 
 | Name              | Short description                                                                                                                                                                                            | 🚦  |
 | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :-: |
-| [multiprocess]    | A fork of [multiprocessing] which uses `dill` instead of `pickle` to allow serializing wider range of object types including nested / anonymous functions. We've found this rather more simple to work with. | 🟢  |
+| [multiprocess]    | A fork of [multiprocessing] which uses `dill` instead of `pickle` to allow serializing wider range of object types including nested / anonymous functions. We've found this easier to use than `multiprocessing`. | 🟢  |
 | [dask]            | Aims to make scaling existing code in familiar libraries (`numpy`, [pandas], `scikit-learn`, ...) easy.                                                                                                      | 🟠  |
 | [multiprocessing] | The standard library module for distributing tasks across multiple processes.                                                                                                                                | 🟠  |
 | [mpi4py]          | Support for MPI based parallelism.                                                                                                                                                                           | 🟠  |
--- a/docs/pages/parallel-async.md
+++ b/docs/pages/parallel-async.md
@@ -13,15 +13,22 @@ reasons to use a specific interface or parallelisation scheme. Possibly due to
 the nature of the research problem, the high-performance computing resources
 available or simply due to pre-existing code using a library like [pandas].
 
+## Thread- and process-based parallelism
+
 | Name              | Short description                                                                                                                                                                                            | 🚦  |
 | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :-: |
 | [multiprocess]    | A fork of [multiprocessing] which uses `dill` instead of `pickle` to allow serializing wider range of object types including nested / anonymous functions. We've found this rather more simple to work with. | 🟢  |
-| [multiprocessing] | The standard library module for distributing tasks across multiple processes                                                                                                                                 | 🟠  |
-| [Cython]          | Has [support for OpenMP based parallelism](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html)                                                                                           | 🟠  |
-| [mpi4py]          | support for MPI based parallelism                                                                                                                                                                            | 🟠  |
 | [dask]            | Aims to make scaling existing code in familiar libraries (`numpy`, [pandas], `scikit-learn`, ...) easy.                                                                                                      | 🟠  |
-| [numba]           | [Support for parallelism via `jit(parallel=True)`](https://numba.pydata.org/numba-doc/latest/user/parallel.html).                                                                                            | 🟠  |
-| [jax]             | [Support for parallelising NumPy / scientific computing like operations using functional transforms](https://jax.readthedocs.io/en/latest/jax-101/06-parallelism.html).                                      | 🟠  |
+| [multiprocessing] | The standard library module for distributing tasks across multiple processes                                                                                                                                 | 🟠  |
+| [mpi4py]          | support for MPI based parallelism                                                                                                                                                                            | 🟠  |
+
+## Compiler-based parallelism
+
+| Name     | Short description                                                                                                                                                       | 🚦  |
+| -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :-: |
+| [Cython] | Has [support for OpenMP based parallelism](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html)                                                      | 🟠  |
+| [numba]  | [Support for parallelism via `jit(parallel=True)`](https://numba.pydata.org/numba-doc/latest/user/parallel.html).                                                       | 🟠  |
+| [jax]    | [Support for parallelising NumPy / scientific computing like operations using functional transforms](https://jax.readthedocs.io/en/latest/jax-101/06-parallelism.html). | 🟠  |
 
 <!-- URLs for more a readable table and text above 👆 -->
 
--- a/docs/pages/parallel-async.md
+++ b/docs/pages/parallel-async.md
@@ -39,7 +39,7 @@ simply due to pre-existing code using a library like [pandas].
 
 <!-- URLs for more a readable tables and text above 👆 -->
 
-[multiprocess]: https://multiprocess.readthedocs.io/en/latest/
+[multiprocess]: https://multiprocess.readthedocs.io/en/stable/
 [multiprocessing]: https://docs.python.org/3/library/multiprocessing.html
 [threading]: https://docs.python.org/3/library/threading.html
 [PEP703]: https://peps.python.org/pep-0703/
--- a/docs/pages/parallel-async.md
+++ b/docs/pages/parallel-async.md
@@ -5,22 +5,29 @@ layout: default
 
 # Parallel and asynchronous processing
 
-Python has a good ecosystem of libraries for multiprocessing (threads and GPU
-parallelisation), as well as asynchronous processing.
+Python has a good ecosystem of libraries for parallelising the processing of tasks,
+as well as asynchronous processing.
 
-🟠 tools in the following may be preferred over 🟢, if there are external
-reasons to use a specific interface or parallelisation scheme. Possibly due to
-the nature of the research problem, the high-performance computing resources
-available or simply due to pre-existing code using a library like [pandas].
+Parallelisation in Python is typically _process-based_ with code parallelised
+across multiple Python processes each with their own interpreter or makes use of
+tools which run the tasks to be parallelised outside of the Python interpreter,
+using for example Python wrappers around external code which uses thread-based
+parallelism.
 
-## Thread- and process-based parallelism
+🟠 tools in the following should be chosen, if there are external reasons to use
+a specific interface or parallelisation scheme. Possibly due to the nature of
+the research problem, the high-performance computing resources available or
+simply due to pre-existing code using a library like [pandas].
+
+## Process-based (and thread-based) parallelism
 
 | Name              | Short description                                                                                                                                                                                            | 🚦  |
 | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :-: |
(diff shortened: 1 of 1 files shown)


docs/pages/parallel-async.md
```diff
@@ -0,0 +1,35 @@
+---
+title: Parallel and asynchronous processing
+layout: default
+---
+
+Python has a good ecosystem of libraries for multiprocessing (threads and GPU
+parallelisation), as well as asynchronous processing. Here, we list those that
+we've found to be useful, particularly for research applications and previous
+ARC projects.
```
matt-graham:
```suggestion
Python has a good ecosystem of libraries for parallelising the processing of tasks,
as well as asynchronous processing. While Python has built-in support for multithreading via 
[the `threading` module in the standard library](https://docs.python.org/3/library/threading.html)
the _global interpreter lock_ (GIL) [currently](https://peps.python.org/pep-0703/) 
means that only one thread can execute Python code at a time. Parallelisation in Python 
therefore typically either is _process-based_ with code parallelised across multiple Python 
processes each with their own interpreter (and GIL) or makes use of tools which run the 
tasks to be parallelised outside of the Python interpreter, using for example Python 
wrappers around external code which uses thread-based parallelism. For the latter case,
some tools use code generation and just-in-time compilation to allow writing code in a 
Python like syntax while still supporting native thread based parallelism. Increasingly there
is also support for parallelising operations on accelerator devices such as graphics 
processing units.
 
Here, we list Python tools that offer support for parallelisation that
we've found to be useful, particularly for research applications and previous
ARC projects.
```

I think it might be worth adding some comment about `threading` / support for thread-based parallelism in Python as it might otherwise seem odd why we don't recommend using the `threading` module for someone who is not familiar with GIL (which is technically a CPython implementation detail so some of my edits above should say CPython instead of Python but I assumed we don't want to go down this whole different rabbit hole if trying to keep things relatively approachable for beginners). I suspect my edits above probably have too much technical detail and might be better in a comment after the actual recommendations, but I think it is worth mentioning the two main approaces to parallelism and why typically we use process based parallelism rather than threads (which would be the default in most other languages).

docs/pages/parallel-async.md
```diff
@@ -0,0 +1,35 @@
...
+
+🟠 tools in the following may be preferred over 🟢, if there are external
+reasons to use a specific interface or parallelisation scheme. Possibly due to
+the nature of the research problem, the high-performance computing resources
+available or simply due to pre-existing code using a library like [pandas].
+
+| Name              | Short description                                                                                                                                                                                            | 🚦  |
+| ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :-: |
+| [multiprocess]    | A fork of [multiprocessing] which uses `dill` instead of `pickle` to allow serializing wider range of object types including nested / anonymous functions. We've found this rather more simple to work with. | 🟢  |
```
paddyroddy: ❤️ `multiprocess`

paddyroddy (commented):
I understand the philosophy behind the traffic light system, but one can easily use i.e. `numba` and `multiprocess` in conjunction - so not sure if it makes too much sense.

docs/pages/parallel-async.md
```diff
@@ -0,0 +1,35 @@
...
+| [multiprocessing] | The standard library module for distributing tasks across multiple processes                                                                                                                                 | 🟠  |
+| [Cython]          | Has [support for OpenMP based parallelism](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html)                                                                                           | 🟠  |
+| [mpi4py]          | support for MPI based parallelism                                                                                                                                                                            | 🟠  |
```
matt-graham:
```suggestion
| [Cython]          | Has [support for OpenMP based parallelism](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html)                                                                                           | 🟠  |
| [mpi4py]          | Support for _message passing interface_ (MPI) based parallelism                                                                                                                                              | 🟠  |
```
Changing case to be consistent with other rows + think its worth expanding MPI acronym as might not be familiar to some

matt-graham (commented):
Thanks @samcunliffe this looks good. Have made some suggestions about adding a bit more detail around different parallelism models in Python and briefly mentioning global interpreter lock to explain why we can't just use `threading` module. I don't if worth specifically adding a separate section on libraries with support for running code on accelerator devices like GPUs? Of the tools already mentioned JAX and [Numba ](https://numba.pydata.org/numba-doc/latest/cuda/index.html) I think have built-in support for running code on GPUs. Other options similar to JAX would be CuPy, TensorFlow and PyTorch, and there also options like `pycuda`, `cuda-python` and `pyopencl`.

docs/pages/parallel-async.md
dstansby:
```suggestion
parallelisation), as well as asynchronous processing.
```
This is just a repeat of the whole purpose of the package, so worth dropping?

docs/pages/parallel-async.md
dstansby: More simple to work with than what?

dstansby (commented):
Maybe this could be split into thread/process based parallelism (`multiprocess`, `multiprocessing`, `dask`, `mpi4py` I think?) and (JIT) compiler based parallelism (`numba`, `jax`, `Cython`)?

samcunliffe:
@paddyroddy 

> I understand the philosophy behind the traffic light system, but one can easily use i.e. numba and multiprocess in conjunction - so not sure if it makes too much sense.

I think @dstansby 's suggestion

>Maybe this could be split into thread/process based parallelism (multiprocess, multiprocessing, dask, mpi4py I think?) and (JIT) compiler based parallelism (numba, jax, Cython)?

Solves your problem?

docs/pages/parallel-async.md
samcunliffe: Than multiprocessing. The subject of the previous sentence.

docs/pages/parallel-async.md
samcunliffe: Should we put threading in the table as :red_circle: ?

matt-graham:
> Maybe this could be split into thread/process based parallelism (`multiprocess`, `multiprocessing`, `dask`, `mpi4py` I think?) and (JIT) compiler based parallelism (`numba`, `jax`, `Cython`)?

I like this suggestion as I think this both deals with @paddyroddy's comment regarding traffic light recommendation and makes it a bit easier for users to identify relevant tool for a specific context. I guess we might then have to recommend a specific tool for compiler based parallelism which might require some thought / discussion though.

docs/pages/parallel-async.md
samcunliffe: Less words is less words 👍

docs/pages/parallel-async.md
samcunliffe: Could just drop the second sentence. Is it clearer that way?

paddyroddy:
> Solves your problem?

Agreed

paddyroddy:
Worth mentioning https://docs.python.org/3/library/concurrent.futures.html too

samcunliffe:
@paddyroddy 

> Worth mentioning https://docs.python.org/3/library/concurrent.futures.html too

I'm gonna need a traffic light colour for that, old bean.

paddyroddy:
> I'm gonna need a traffic light colour for that, old bean.

I'm a fan! 🎐

paddyroddy:
@matt-graham GIL _may_ go in the future https://peps.python.org/pep-0703

matt-graham:
> @matt-graham GIL _may_ go in the future https://peps.python.org/pep-0703

Yep I linked to that PEP in my suggested change above 😉 

EDIT: Potentially misunderstood your comment, was the point that is not certain and so not worth linking to?

paddyroddy:
> Yep I linked to that PEP in my suggested change above 😉
> 
> EDIT: Potentially misunderstood your comment, was the point that is not certain and so not worth linking to?

To be honest, it is so many words that my brain skipped some of it... I feel like it is going into a lot of detail that won't be relevant for most users. However, perhaps it could be useful in a `<details>` block?

But yes it is also may fall out of date with the Python ecosystem, so may just be worth linking to the PEP?

docs/pages/parallel-async.md
matt-graham: Yeah that could work as a more succinct way of getting across the same point

samcunliffe:
>  I guess we might then have to recommend a specific tool for compiler based parallelism which might require some thought / discussion though.

Left them as all-amber for now. I'd be inclined to factor out the discussion of our favorites to either a new ticket, or to an in-person meetup.

docs/pages/parallel-async.md
samcunliffe: [1f8f261](https://github.com/UCL-ARC/python-tooling/pull/323/commits/1f8f26132ef85eee197cbc0fa646cb8a91d329b2)

paddyroddy:
> Worth mentioning [docs.python.org/3/library/concurrent.futures.html](https://docs.python.org/3/library/concurrent.futures.html) too

Part of standard library too

docs/pages/parallel-async.md
dstansby:
```suggestion
| [multiprocess]    | A fork of [multiprocessing] which uses `dill` instead of `pickle` to allow serializing wider range of object types including nested / anonymous functions. We've found this easier to use than `multiprocessing`. | 🟢  |
```

docs/pages/parallel-async.md
```diff
@@ -0,0 +1,51 @@
...
+
+| Name     | Short description                                                                                                                                                       | 🚦  |
+| -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :-: |
+| [Cython] | Has [support for OpenMP based parallelism](https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html)                                                      | 🟠  |
+| [numba]  | [Support for parallelism via `jit(parallel=True)`](https://numba.pydata.org/numba-doc/latest/user/parallel.html).                                                       | 🟠  |
+| [jax]    | [Support for parallelising NumPy / scientific computing like operations using functional transforms](https://jax.readthedocs.io/en/latest/jax-101/06-parallelism.html). | 🟠  |
+
+<!-- URLs for more a readable tables and text above 👆 -->
+
+[multiprocess]: https://multiprocess.readthedocs.io/en/latest/
```
dstansby:
```suggestion
[multiprocess]: https://multiprocess.readthedocs.io/en/stable/
```

samcunliffe: 😱

dstansby (approved):

matt-graham:
> To be honest, it is so many words that my brain skipped some of it... 

I wouldn't say ~150 words is all that many, and irrespective I feel it's not unreasonable to expect you to read things you comment on before commenting on them, even if you find them overly wordy 🙁. Not trying to make a big deal of this, but comments like the above do make it less appealing to contribute and I would say we should keep comments kind and constructive.

> I feel like it is going into a lot of detail that won't be relevant for most users. However, perhaps it could be useful in a `<details>` block?
> 
> But yes it is also may fall out of date with the Python ecosystem, so may just be worth linking to the PEP?

Even if/when PEP0703 gets fully implemented the GIL will be _optional_ but not removed so a lot of the mentioned details will still be relevant. I think having some explanatory comments is useful as while we want to keep thing simple for users, wanting to exploit parallelism / asynchronicity is a relatively advanced use case, and is inherently a complex topic and I think trying to hide away the complexity doesn't necessarily help. Putting some extra detail in a `<details>` block I think would be a good approach.

paddyroddy:
> I wouldn't say ~150 words is all that many, and irrespective I feel it's not unreasonable to expect you to read things you comment on before commenting on them, even if you find them overly wordy 🙁. Not trying to make a big deal of this, but comments like the above do make it less appealing to contribute and I would say we should keep comments kind and constructive.

I did read them... What I am saying is that is why I missed you'd already linked to the issue. We all have different reading and interpreting styles.

docs/pages/parallel-async.md
```diff
@@ -0,0 +1,51 @@
...
+# Parallel and asynchronous processing
+
+Python has a good ecosystem of libraries for parallelising the processing of tasks,
+as well as asynchronous processing.
+
+Parallelisation in Python is typically _process-based_ with code parallelised
+across multiple Python processes each with their own interpreter or makes use of
+tools which run the tasks to be parallelised outside of the Python interpreter,
+using for example Python wrappers around external code which uses thread-based
+parallelism.
```
samcunliffe, line 15: @matt-graham 

Here ☝️  (and l30 👇 ) is what I settled on, but I don't go into detail about the GIL. Do you think we need a `<details>` for the rest?

> Even if/when PEP0703 gets fully implemented the GIL will be optional but not removed so a lot of the mentioned details will still be relevant. I think having some explanatory comments is useful as while we want to keep thing simple for users, wanting to exploit parallelism / asynchronicity is a relatively advanced use case, and is inherently a complex topic and I think trying to hide away the complexity doesn't necessarily help. Putting some extra detail in a `<details>` block I think would be a good approach.

samcunliffe:
To stop this PR snowballing, I'm factoring out the asynchronous stuff to
- #331

docs/pages/parallel-async.md
matt-graham: Thanks @samcunliffe for the updates - the extra detail in the intro paragraph plus the inclusion of `threading` module in 🚦 table with a reference to GIL there is I think works well, so no need for a `<details>` tag.

matt-graham (approved):

PR #324, opened 2024-03-23
Add packaging tips.

### LINKED ISSUE #103: Clarify some (obvious to us) best practices about repo organisation etc.
We probably want somewhere to write down, i.e. one package per repo. And link to package vs module distinction (probably somewhere in the official python docs).

These kinds of things are obvious to us, but likely not to new python developers.

Not really sure _where_ this should go. But probably should be somewhere here.

### Relates to

* https://github.com/UCL-ARC/python-template/discussions/53

samcunliffe:
Solves #103.

@willGraham01 from your work on brainglobe, perhaps you have some insights or lessons learned that might influence this text?

### DIFF
--- a/docs/pages/packaging.md
+++ b/docs/pages/packaging.md
@@ -10,9 +10,10 @@ package?_". Where code to solve a research problem might be large and perform
 several tasks at once typically one should try to stick to a mantra of doing one
 thing well.
 The following questions can be helpful when trying to abide by this mantra:
+
 - What do I want my package to provide?
 - How will users interact with or use the code in my package?
-- Is everything I'm including in the package relevant or useful in supporting it's main purpose? 
+- Is everything I'm including in the package relevant or useful in supporting it's main purpose?
 
 In Python, a [package](https://docs.python.org/3/reference/import.html#packages)
 is a collection of one or more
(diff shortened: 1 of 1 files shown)


docs/pages/packaging.md
```diff
@@ -3,7 +3,22 @@ title: Packaging
 layout: default
 ---
 
-# Packaging
+# General packaging
+
+A common question facing developers is "_How much code should go into a
+package?_". Where code to solve a research problem might be large and perform
+several tasks at once typically one should try to stick to a mantra of doing one
+thing well.
```
willGraham01, line 11: Probably having some idea of scope before you start writing your package would be my main piece of advice from BrainGlobe, though you've kind of hinted at that in the new text already. Could maybe add

```suggestion
several tasks at once typically one should try to stick to a mantra of doing one
thing well.
The following questions can be helpful when trying to abide by this mantra:
- What do I want my package to provide?
- How will users interact with or use the code in my package?
- Is everything I'm including in the package relevant or useful in supporting it's main purpose? 
```

paddyroddy (approved):

matt-graham (approved):
Looks good to me. I like @willGraham01's suggested expansion

PR #326, opened 2024-03-23
Revert "Correct uses of project_name in pyproject.toml"

dstansby:
Reverts UCL-ARC/python-tooling#309. It turns out this was incorrect... source: a new package I just made. maybe at some point we should add some regression tests for the contents of `pyproject.toml`?

### DIFF
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -49,7 +49,7 @@ urls.homepage = "https://github.com/{{cookiecutter.github_username}}/{{cookiecut
 [tool.coverage]
 report = {sort = "cover"}
 run = {branch = true, parallel = true, source = [
-    "{{cookiecutter.project_name}}",
+    "{{cookiecutter.project_slug}}",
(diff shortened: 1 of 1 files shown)


samcunliffe (approved):
The old module/package/slug confusion? Yeah, it's a bit disturbing that we didn't have a test that failed for #309. 

> add some regression tests for the contents of pyproject.toml?

Are you thinking just a simple diff with an expected output? We could try to test some edge cases, at least (different project name w.r.t. slug with spaces).

paddyroddy:
> Reverts #309. It turns out this was incorrect... source: a new package I just made. maybe at some point we should add some regression tests for the contents of `pyproject.toml`?

I was surprised when you said this wasn't working! But blindly trusted, so yes we need a test.

dstansby:
> Are you thinking just a simple diff with an expected output?

Yes, but I think worth merging this first to un-break the templating, and I can open a new issue for better tests.

dstansby:
Issue opened at https://github.com/UCL-ARC/python-tooling/issues/329

PR #328, opened 2024-03-24
Start a utility libraries page.

### LINKED ISSUE #304: Implement some kind of `logging`
Either something like this https://github.com/astro-informatics/sleplet/blob/6418525d4f8bb0b521205d5c2471125ac3172cde/src/sleplet/__init__.py#L30-L40 or a package which does the same thing

samcunliffe:
Start a utilities page rather than solving #304 as-requested.

As I said in the ticket: I'm 👎 about adding these kinds of dependencies to our template. Not least because not everyone _needs_ a logger, but also because these choices feel out-of-scope.  We probably _should_ highlight libraries that we like and prefer to the builtins.

I've added my opinions, such as they are. Challenges are welcome.

## Closes
- #304.

### DIFF
--- a/docs/pages/utilities.md
+++ b/docs/pages/utilities.md
@@ -1,0 +1,26 @@
+---
+title: Utility libraries
+layout: default
+---
+
+# Utility libraries
+
+Useful packages for 'utility' functionality, normally we highlight these if
+we've found a library that we prefer to Python's builtin. You probably don't
+want to write these yourself but might need logging or a command-line interface.
+
+## Command-line interface
+
+| Name                                                        | Short description                                                              | 🚦  |
+| ----------------------------------------------------------- | ------------------------------------------------------------------------------ | :-: |
+| [typer](https://typer.tiangolo.com/)                        | A user-friendly and intuitive CLI system. Built on click.                      | 🟢  |
+| [click](https://click.palletsprojects.com/)                 | Makes nice-looking CLIs, and more configurable than argparse. Uses decorators. | 🟠  |
+| [argparse](https://docs.python.org/3/library/argparse.html) | Python's builtin CLI system uses object configuration.                         | 🟠  |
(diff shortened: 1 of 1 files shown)


docs/pages/utilities.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Utility libraries
+layout: default
+---
+
+# Utility libraries
+
+Useful packages for 'utility' functionality, normally we highlight these if
```
paddyroddy:
```suggestion
Useful packages for _utility_ functionality, normally we highlight these if
```

docs/pages/utilities.md
```diff
@@ -0,0 +1,26 @@
...
+want to write these yourself but might need logging or a command-line interface.
+
+## Command-line interface
+
+| Name                                                        | Short description                                                              | 🚦  |
+| ----------------------------------------------------------- | ------------------------------------------------------------------------------ | :-: |
+| [typer](https://typer.tiangolo.com/)                        | A user-friendly and intuitive CLI system. Built on click.                      | 🟢  |
+| [click](https://click.palletsprojects.com/)                 | Makes nice-looking CLIs, and more configurable than argparse. Uses decorators. | 🟠  |
+| [argparse](https://docs.python.org/3/library/argparse.html) | Python's builtin CLI system uses object configuration.                         | 🟠  |
+| [optparse](https://docs.python.org/3/library/optparse.html) | A now-deprecated CLI system built into Python.                                 | 🔴  |
```
paddyroddy, line 19: If it's deprecated, why have it?

samcunliffe: Because it still exists (it's possible to `import optparse` ) we want people to _know_ not to use it.

>  We actively discourage using this. This could be because it’s no longer maintained, not open source, or difficult to use. Consider moving to alternatives if you’re currently using something that’s red. A reason for not using this is given.

paddyroddy (approved):
🫐

PR #331, opened 2024-03-25
Add asynchronous processing libraries.

samcunliffe:
The first couple of async tools. I don't have tonnes of experience with async stuff so I'm probably not the best person to decide the 🚦 s.

We use `asyncio` for [INFORMus](https://github.com/inform-us/INFORMus/) (and probably a fair few other web APIs??) and @paddyroddy likes `futures`.

## Continues

- #323 
- #178

### DIFF
--- a/docs/pages/parallel-async.md
+++ b/docs/pages/parallel-async.md
@@ -41,8 +41,8 @@ simply due to pre-existing code using a library like [pandas].
 
 | Name                 | Short description                                                                                                                                 | 🚦  |
 | -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- | :-: |
-| [asyncio]            | The Python standard library for simple asynchronous processing. Useful for simple cases.                                                          | 🟠  |
-| [concurrent.futures] | Another Python standard library for more complex asynchrounous processing, comes with a bigger overhead but better suited to data analysis tasks. | 🟠  |
+| [asyncio]            | Python standard library for asynchronous programming with tasks run in a single-threaded event loop. Used for cooperative multitasking](https://en.wikipedia.org/wiki/Cooperative_multitasking).            | 🟠  |
+| [concurrent.futures] | Another Python standard library for asynchrounous processing. Provides a common interface for thread and process based concurrency as an alternative to using `multiprocess(ing)` or `threading` directly.  | 🟠  |
 
 <!-- URLs for more a readable tables and text above 👆 -->
 
--- a/docs/pages/parallel-async.md
+++ b/docs/pages/parallel-async.md
@@ -5,8 +5,8 @@ layout: default
 
 # Parallel and asynchronous processing
 
-Python has a good ecosystem of libraries for parallelising the processing of tasks,
-as well as asynchronous processing.
+Python has a good ecosystem of libraries for parallelising the processing of
+tasks, as well as asynchronous processing.
 
 Parallelisation in Python is typically _process-based_ with code parallelised
 across multiple Python processes each with their own interpreter or makes use of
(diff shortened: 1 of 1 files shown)


docs/pages/parallel-async.md
```diff
@@ -37,6 +37,13 @@ simply due to pre-existing code using a library like [pandas].
 | [numba]  | [Support for parallelism via `jit(parallel=True)`](https://numba.pydata.org/numba-doc/latest/user/parallel.html).                                                       | 🟠  |
 | [jax]    | [Support for parallelising NumPy / scientific computing like operations using functional transforms](https://jax.readthedocs.io/en/latest/jax-101/06-parallelism.html). | 🟠  |
 
+## Asynchronous processing
+
+| Name                 | Short description                                                                                                                                 | 🚦  |
+| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- | :-: |
+| [asyncio]            | The Python standard library for simple asynchronous processing. Useful for simple cases.                                                          | 🟠  |
+| [concurrent.futures] | Another Python standard library for more complex asynchrounous processing, comes with a bigger overhead but better suited to data analysis tasks. | 🟠  |
```
matt-graham: My understanding is that `asyncio` and `concurrent.futures` tackle slightly different problems, even though there is overlap between them. 

`asyncio` provides one way of writing co-routines in Python to allow co-operative multitasking, as well an 'engine' for running co-routines concurrently and managing one or more event loops. The `async` / `await` syntax for defining co-routines that has been available since Python 3.5 can be also used to define co-routines to run with [event-loops provided by other libraries such as `tornado`](https://www.tornadoweb.org/en/stable/guide/async.html). The event-loop abstraction makes `asyncio` (and related libraries) particularly useful in client / server and graphical user interface settings. 

`asyncio` can also be useful in other situations - for example dealing with communication with subprocesses. We used `asyncio` in AQUIFER to allow running [asynchronous dispatch of runs of a simulator in a subprocess](https://github.com/UCL/neso-calibration/blob/main/src/nesopy/model.py) so we could both continue running other Python code while the model is running in the background and also have any model output to `stdout`  / `stderr` redirected to the parent process while running rather than only getting access to the output on the subprocess exiting.

`concurrent.futures` on the other hand, I think is more of an abstraction / interface layer over different approaches for harnessing concurrency in Python, in particular allowing you to easily swap between thread-based and process-based approaches. It therefore in some ways more of a competitor to `multiprocess(ing)` and `threading` modules.

[This StackOverflow answer](https://stackoverflow.com/a/61360215) I think is a nice summary of what each of `threading`, `multiprocessing`, `asyncio` and `concurrent.futures` do.

paddyroddy: Yeah, the way I've used `concurrent.futures` has been in place of `multiprocess`, i.e. as
```python
with concurrent.futures.ThreadPoolExecutor(max_workers=ncpu) as e:
    e.map(func, chunks)
```
it felt nicer to use something from the standard library, which appears to be slightly more powerful and should retain support.

I was still using `multiprocess.shared_memory.SharedMemory` as `concurrent.futures` didn't have support for that.

samcunliffe: >  It therefore in some ways more of a competitor to multiprocess(ing) and threading modules.

Hmm. Perhaps we put it in both tables? ...And link to the SO post?

matt-graham: Sorry realised this isn't really a helpful comment as it doesn't really say what to change. A suggested rephrasing:

Maybe 
```suggestion
| [asyncio]            | Python standard library for asynchronous programming with tasks run in a single-threaded event loop. Used for cooperative multitasking](https://en.wikipedia.org/wiki/Cooperative_multitasking).            | 🟠  |
| [concurrent.futures] | Another Python standard library for asynchrounous processing. Provides a common interface for thread and process based concurrency as an alternative to using `multiprocess(ing)` or `threading` directly.  | 🟠  |
```

matt-graham: Missed the above replies before posting my updated suggestion. Putting in both tables seems reasonable, and linking to the StackOverflow post I think would be good.

matt-graham (approved):

PR #332, opened 2024-03-25
Minor improvement to template sentence

dstansby:
- The 🍪 is a bit confusing out of context here, so remove
- "template" > "Python package template" as this might be the first time someone reads about a template, and otherwise they might wonder what it's for.


samcunliffe:
> The 🍪 is a bit confusing out of context here, so remove

🍪 🥺

samcunliffe (approved):

paddyroddy (approved):

Issue #310, opened 2024-02-21
Markdown link checker?


sfmig:
### What is the Topic of Your Question

Pre-commit hooks

### Add Your Question Below

Should we add a pre-commit hook that checks links are alive in markdown files?

Something like [this](https://github.com/tcort/markdown-link-check).

I found it useful recently 😬

paddyroddy:
Not saying we shouldn't, but we already have link checking in CI https://github.com/UCL-ARC/python-tooling/blob/main/.github/workflows/links.yml (which is quite flaky)

dstansby:
I think having a link checker would generate too many false negatives (e.g. when a website happens to be down) to be used in a pre-commit hook, which is run on every commit. How about adding a tox environment that runs a link check? That way it would be easy to run it manually occasionally.

paddyroddy:
I don't think we need both, we have it in CI. Considering flakiness is expected, I think just having CI (which can be re-run) should be enough.

sfmig:
CI only sounds good!

paddyroddy:
Thanks for the suggestion @sfmig, closing 🦀

PR #335, opened 2024-03-25
Minor improvements to CI page

dstansby:
- Don't put an acronym in table of contents
- Note that GH Actions also available for private repos for free
- Link to testing page for test coverage collection recommendation
- Add a details box explaining why there are two 🟢 coverage hosting tools


docs/pages/ci.md
```diff
@@ -15,19 +15,21 @@ layout: default
...
 
-You can check coverage simply in the terminal with the
-[pytest-cov](https://pypi.org/project/pytest-cov/) plugin. However you might
-consider a third-party code coverage analytics and reporting service. There are
-two that we've used and they're both rather similar. They render the code with
-highlighting to show which lines are not executed but the tests, and can track
-test coverage over time.
+These services report and track test code coverage over time.
+They render the code with highlighting to show which lines are not executed by tests.
+See [testing](testing.md) for our recommendations on packages to generate code coverage during tests.
```
samcunliffe: I claim Jekyll is not smart enough to make this a link.
```
[2024-03-25 17:27:17] ERROR `/pages/testing.md' not found.
```

samcunliffe (changes requested):

docs/pages/ci.md
dstansby: For some reason I thought we were using `mkdocs` so I went and looked up how to link in the mkdocs docs... 🤦

dstansby: Hold on how did you work this out? All the CI is 🟢 , did you do a local build? CI definitely shouldn't be 🟢 if the docs won't build...

samcunliffe: Yeah I just did a `bundle exec jekyll serve` locally... Actually, it's because I would've been impressed that links were replaced like that.

Which might mean I'm impressed with `MkDocs`...? 🤣

samcunliffe: I guess our CI doesn't complain because the `jekyll`/JTD build doesn't do any link-checking/

paddyroddy: #310 think it was decided against link checking in `pre-commit`. However, local link s should be fine.

With `mkdocs` I would do `mkdocs build --strict` which would fail in this case. `jekyll` probably has something similar.

docs/pages/ci.md
```diff
@@ -15,19 +15,21 @@ layout: default
 
 <details>
 <summary> 🟢 explanation</summary>
-We have many projects using github CI, it's both convenient and works well. It's become the standard. Also, note that it's free for public repos.
+We have many projects using GitHub CI and, it has good integration with GitHub itself, and is free for public repositories (with limited free monthly minutes for private repositories).
```
paddyroddy, line 18: Seeing as this is the first time we are using CI, should we define it?
```suggestion
We have many projects using GitHub continuous integration (CI) and, it has good integration with GitHub itself, and is free for public repositories (with limited free monthly minutes for private repositories).
```

paddyroddy: Would be nice if this page could be wrapped to 80 chars, in keeping with other pages

docs/pages/ci.md
samcunliffe: There's `bundle exec jekyll build  --strict_front_matter` but that doesn't seem to fail here.

samcunliffe:
```suggestion
See [testing](testing) for our recommendations on packages to generate code coverage during tests.
```

Seems to work!

docs/pages/ci.md
dstansby: I think if we want to enforce wrap to 80 it would be best to do via a pre-commit hook instead of having to manually edit?

docs/pages/ci.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: CI
+title: Continuous integration
```
dstansby, line 2:
```suggestion
title: Continuous integration (CI)
```

docs/pages/ci.md
dstansby: Thanks for the fix! I opened https://github.com/UCL-ARC/python-tooling/issues/336 to track making the doc build fail.

docs/pages/ci.md
dstansby: Thanks for catching the acronym - I edited the title of this section instead since CI is such a common acronym.

paddyroddy: Sure, it's slightly tricky though as links can go over - and tables. I like [markdownlint](https://github.com/paddyroddy/.github/blob/c3bb562d91f4004231cdd8497106a209a7531f40/precommit/general/general-hooks.yml#L56-L59).

dstansby: For wrapping I opened a new issue at https://github.com/UCL-ARC/python-tooling/issues/337

samcunliffe (approved):

PR #343, opened 2024-03-27
Remove default issue assignees

dstansby:
My feeling is that adding default assignees to issues makes the project less welcoming for new (or existing) people to get involved with - if all issues are assigned I would certainly be wary of opening a PR to address them.


samcunliffe (approved):
Fine by me. IIRC the default assignee was just so tickets don't get accidentally ignored.

paddyroddy (approved):
Yeah, agree with Sam. But happy to receive fewer notifications.

PR #348, opened 2024-04-02
Minor improvements to profiling docs

dstansby:
- Note that `memory_profiler` could be useful on Winows, where memray isn't available
- Make memory_profiler orange (from red before). I don't think we should actively discourage using it, even if it is unmaintained at the moment it can still be useful (especially on Windows)
- Split profiling tools up into sub-sections


paddyroddy (commented):
Is `psutil` meant to be there twice?

dstansby:
Woops, I think one of those entries was meant to be `pyinstrument`. Should be fixed now.

paddyroddy (approved):
🫐

PR #358, opened 2024-04-25
Open link in new tab

paddyroddy:


paddyroddy:
Error is to do with https://github.com/actions/setup-python/issues/850

dstansby:
We don't have any way of viewing rendered docs from a PR to check them before merging do we?

paddyroddy:
> We don't have any way of viewing rendered docs from a PR to check them before merging do we?

I don't believe so

samcunliffe:
Doesn't work?

https://github.com/UCL-ARC/python-tooling/assets/1836192/38e623a5-7b23-4d2b-adb2-12ae4ff16ada


```
➤ git status
On branch paddyroddy-patch-1
Your branch is up to date with 'origin/paddyroddy-patch-1'.
```

paddyroddy:
> Doesn't work?

Click on `Source repository` top right

paddyroddy:
Individual links have to be handled in a different manner

samcunliffe:
I like the idea, but I'd go with consistency. Only having the source link open in a new tab is... 🤷 meh. All external links? I could see the benefit.

paddyroddy:
> I like the idea, but I'd go with consistency. Only having the source link open in a new tab is... 🤷 meh. All external links? I could see the benefit.

But that would require going through every single link I believe, i.e. a different PR

giordano:
> We don't have any way of viewing rendered docs from a PR to check them before merging do we?

For what is worth, we do that in the excalibur benchmarking repository: https://github.com/ukri-excalibur/excalibur-tests/blob/b53e6600c827ae797183c37edc5f7b598439b08c/.github/workflows/docs.yml (this is using mkdocs rather than jekyll, but to give an idea of what one can do to accomplish that).

dstansby (approved):

PR #360, opened 2024-04-25
Drop `3.9` support

### LINKED ISSUE #359: Change CI to either use `macos-13` or drop `3.9` support
### Describe the Bug

See bug described here https://github.com/actions/setup-python/issues/850, essentially `macos-latest` now represents ARM mac 😱

### To Reproduce

Use `macos-latest` in conjunction with python versions `<3.10`

### Expected Behaviour

-

### Actual Behaviour

-

### Additional Context

```Markdown
- Cookiecutter version:
- Operating system: macos
- Python version: 3.9
```

paddyroddy:
Following the recent change on GitHub Actions https://github.com/actions/setup-python/issues/850 our tests are currently failing https://github.com/UCL-ARC/python-tooling/issues/359.

Previously we were following [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) which is why we never supported `3.8`. My suggestion is to drop `3.9` which means nothing else has to change.

The alternative is to do @dstansby's [suggestion](https://github.com/UCL-ARC/python-tooling/issues/359#issuecomment-2077127044), but I think this is low-key complicated when involving our `tox` setup too. Personally, I don't think it's worth it, and we should fix the tests ASAP.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,6 @@ jobs:
           - ubuntu-latest
           - windows-latest
         python-version:
-          - "3.9"
           - "3.10"
           - "3.11"
           - "3.12"
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -9,8 +9,8 @@
   "initialise_git_repository": true,
   "deploy_docs_to_github_pages": false,
   "github_username": "{{cookiecutter.author_given_names.lower().replace(' ', '-')}}-{{cookiecutter.author_family_names.lower().replace(' ', '-')}}",
-  "min_python_version": ["3.9", "3.10", "3.11", "3.12"],
-  "max_python_version": ["3.12", "3.11", "3.10", "3.9"],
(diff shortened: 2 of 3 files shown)


samcunliffe:
I'm 👍 👎 about this.

I agree with the NEP argument. But some non-numpy projects could presumably run just fine with 3.9. I'll go with the majority.

_Edit: [what dstansby said](https://github.com/UCL-ARC/python-tooling/issues/359#issuecomment-2077248489)._

dstansby (changes requested):
I'm personally 👎 on dropping this (it wouldn't be that much more effort to support 3.8/3.9), but happy to go with a consesus the other way. I agree with https://github.com/UCL-ARC/python-tooling/issues/359#issuecomment-2077351878, in that we should come to a consensus decision to choose to follow either SPEC0 or NEP29 before making this change.

dstansby (approved):

Issue #354, opened 2024-04-08
Add section to docs on GUI toolkits


dstansby:
### Is Your Feature Request Related to a Problem? Please Describe

n/a

### Describe the Solution You'd Like

A seciton in the docs on toolkits for making GUIs with Python.

p-j-smith:
Any thoughts on which toolkits to include / recommend? Some we could include:

- [QtPy](https://github.com/spyder-ide/qtpy)
- [PySide6](https://wiki.qt.io/Qt_for_Python)
- [magicgui](https://github.com/pyapp-kit/magicgui)

Plus more application-specific ones:

- [napari](https://github.com/napari/napari)
- [PyVistaQt](https://github.com/pyvista/pyvistaqt)
- [Trame](https://github.com/Kitware/trame) (a web framework for VTK but can also be used to create apps that run locally in the browser)

And should we include packages for creating applications (perhaps that should be a separate section though):

- [PyInstaller](https://github.com/pyinstaller/pyinstaller)
- [pyapp](https://github.com/pyapp-org/pyapp)

samcunliffe:
* [PyQt](https://wiki.python.org/moin/PyQt) vs PySide vs [QtPy](https://pypi.org/project/QtPy/)
  > QtPy is a small abstraction layer that lets you write applications using a single API call to either PyQt or PySide.
  >
  > It provides support for PyQt5, PySide2, PyQt6 and PySide6 using the Qt5 layout (where the QtGui module has been split into QtGui and QtWidgets).

p-j-smith:
> QtPy is a small abstraction layer that lets you write applications using a single API call to either PyQt or PySide.
> 
> It provides support for PyQt5, PySide2, PyQt6 and PySide6 using the Qt5 layout (where the QtGui module has been split into QtGui and QtWidgets).

Yep, QtPy is the way to go you're e.g. developing an app with a plugin system (so other's can choose which Qt bindings to use), or if you're developing a widget that others can use in their apps.

But if you're developing an app that you'll build with PyInstaller/pyapp and distribute as an executable, there might be no need to support multiple bindings and you could choose to use PySide6 directly (or any other bindings). Although I would probably avoid using PyQt over PySide - PyQt has a more restrictive license, PySide is developed by Qt, and the PySide docs are (nowadays) better than the PyQt docs

PR #373, opened 2024-05-09
Start a GUI page

dstansby:
A start on https://github.com/UCL-ARC/python-tooling/issues/354. I don't have recent experience with other GUI toolkits, so leave it for others to fill them in.

I would suggest merging this PR as a starting point, and then iterate on it to add more content in future PRs?


docs/pages/guis.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Graphical user interface toolkits
+
+| Name      | Short description                                                                     | 🚦  |
+| --------- | ------------------------------------------------------------------------------------- | :-: |
+| [tkinter] | A Python interface to the Tcl/Tk GUI toolkit. Built into the Python standard library. | 🟠  |
+
+<!-- URLS used above -->
+
+[tkinter]: https://docs.python.org/3/library/tkinter.html
```
samcunliffe:
```suggestion
| [QtPy]    | A Python interface to the [Qt] GUI system. Recommended if you need to be binding-independent. | 🟢 |
| [PySlide] | A Python binding to the [Qt] GUI system. Recommended if you're developing an app that you'll build and distribute as an executable. In which case you don't need to be binding-independent.  | 🟢 |
| [PyQt] | A Python binding to the [Qt] GUI system. A specific binding but with a more restrictive license.  | 🟠  |
| [tkinter] | A Python interface to the Tcl/Tk GUI toolkit. Built into the Python standard library. | 🟠  |
<!-- URLS used above -->

[tkinter]: https://docs.python.org/3/library/tkinter.html
[Qt]: https://www.qt.io/
[QtPy]: https://github.com/spyder-ide/qtpy
[PySlide]: https://wiki.qt.io/Qt_for_Python
[PyQt]: https://wiki.python.org/moin/PyQt
```

samcunliffe: Taken from @p-j-smith's comment
- https://github.com/UCL-ARC/python-tooling/issues/354#issuecomment-2051416122

samcunliffe:
I'm still 👎 about a different page for each type of library and think we should have a large(ish) single page with all libraries that we recommend/use and like in our projects.

dstansby:
> I'm still 👎 about a different page for each type of library and think we should have a large(ish) single page with all libraries that we recommend/use and like in our projects.

👍 for this in theory (especially as the list of recommendations grows), but deserves it's own issue/PR?

nicolin:
I have just joined and use Qt C++ for GUI development, so a GUI page should sit outside python tooling as a GUI should be solver code agnostic just providing input to the solver and also reading back as xml or text format.

samcunliffe:
> I have just joined and use Qt C++ for GUI development, so a GUI page should sit outside python tooling as a GUI should be solver code agnostic just providing input to the solver and also reading back as xml or text format.

The scope of this is research software projects in general. We are not just for solvers. I think there is enough Python-specific interest in GUI programming and enough Python GUI libraries that the page should exist for signposting.

samcunliffe:
... although we should do something with this PR. 
Merge as-is or move the table to utilities? I favor the latter but I'd bow to the consensus.

dstansby:
I'm 👍 to as is, as I think having a GUI page makes it more discoverable.

paddyroddy:
> I have just joined and use Qt C++ for GUI development, so a GUI page should sit outside python tooling as a GUI should be solver code agnostic just providing input to the solver and also reading back as xml or text format.

I second Sam. The purpose of this project is not to imply that GUI means only for Python. We're saying in the context of Python development, if you want advice on GUIs then here are some potential ideas, and our experience.

p-j-smith:
> > I'm still 👎 about a different page for each type of library and think we should have a large(ish) single page with all libraries that we recommend/use and like in our projects.
>
> 👍 for this in theory (especially as the list of recommendations grows), but deserves it's own issue/PR?
>
I'm in two minds about this, but probably agree that having a single page for these things would be good. Should the libraries in [Parallel and asychronous processing](https://github-pages.arc.ucl.ac.uk/python-tooling/pages/parallel-async.html) also be moved to this `libraries` page? Agree with @dstansby that having a separate issue / PR to discuss this makes sense

samcunliffe:
Merge as-is then?

p-j-smith (approved):
looks good - shall we merge this and continue discussion around a single libraries page in https://github.com/UCL-ARC/python-tooling/issues/427?

Issue #363, opened 2024-04-25
Viewing rendered docs from a PR to check


paddyroddy:
> We don't have any way of viewing rendered docs from a PR to check them before merging do we?

For what is worth, we do that in the excalibur benchmarking repository: https://github.com/ukri-excalibur/excalibur-tests/blob/ede38fa104eed72557233cbfc76b230727973791/.github/workflows/docs.yml (this is using mkdocs rather than jekyll, but to give an idea of what one can do to accomplish that).

_Originally posted by @giordano in https://github.com/UCL-ARC/python-tooling/issues/358#issuecomment-2077433013_

dstansby:
https://readthedocs.org has pretty good support for pull request previews: https://docs.readthedocs.io/en/stable/pull-requests.html

dstansby:
Or https://github.com/marketplace/actions/deploy-pr-preview might be useful, since I don't think readthedocs can (easily) build jekyll sites like this one.

PR #377, opened 2024-06-01
Revert "Lock file maintenance (#376)"

samcunliffe:
This reverts commit PR #376 (commit 2e196ec7f9c431454b6966384c1e09acfae4e6bb) which broke [the pages action](https://github.com/UCL-ARC/python-tooling/actions/runs/9329000737/job/25680916032).

Side note, we should probably rerun the pages action (but without deployment) as a CI check if anything in the docs folder is changed. (A step along the way to #363.)


dstansby (approved):

paddyroddy:
We can remove this line if we don't like that option @samcunliffe https://github.com/UCL-ARC/python-tooling/blob/07639179b43ba27d267dbd7d62e80be74f01757a/.renovaterc.json5#L5 (still experimenting)

PR #405, opened 2024-06-07
Refactor tests

dstansby:
This makes the tests a bit easier to extend by:

- Defining a helper ~~function~~ **_fixture_** to generate a package from a given set of configuration options (in ~~`tests/helpers.py`~~ `conftest.py`)
- Defining the configuration within the tests that are being used, instead of in `conftest.py`. This makes it easier to see what config each test is using next to the actual test.


samcunliffe (approved):
Looks nice!
:blueberries:

samcunliffe:
Ah. The conflicts are from me in 
- #412 

(Soz)

paddyroddy (commented):
Personally, I like a `conftest.py` and they are very commonly used. Might be better to document what it is, and how it works.

tests/helpers.py
```diff
@@ -0,0 +1,30 @@
+"""Checks that the cookiecutter works."""
+
+import pathlib
+import subprocess
+
+
+def gen_package(
```
samcunliffe: Any benefit for this being a fixture ?
(Just a thought. I like it either way.)

samcunliffe (dismissed):

tests/test_git_init.py
```diff
@@ -5,46 +5,28 @@
 
 import pytest
 
+from .helpers import gen_package
+
 
 @pytest.mark.parametrize(
     "git_init_cookiecutter_option",
-    ["initialise_git_repository=True", "initialise_git_repository=False"],
+    ["True", "False"],
```
paddyroddy: I don't like random `True`/`False`

matt-graham (changes requested):
At the moment the approach of making the `tests` directory a package by adding an `__init__.py` (presumably so we can use a relative import for `helpers` module) seems to break the ability to run tests directly by running `pytest` or `pytest -s` from the root of the repository (with the latter being what we currently recommend in the developer notes in the README), though tests do still run fine using `tox`. Specifically I get the following errors on test collection
```
_______________________________________ ERROR collecting tests/test_git_init.py ________________________________________
ImportError while importing test module '.../python-tooling/tests/test_git_init.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_git_init.py:8: in <module>
    from .helpers import gen_package
E   ModuleNotFoundError: No module named 'tests'
______________________________________ ERROR collecting tests/test_package_gen.py ______________________________________
ImportError while importing test module '.../python-tooling/tests/test_package_gen.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_package_gen.py:6: in <module>
    from .helpers import gen_package
E   ModuleNotFoundError: No module named 'tests'
```
I guess this is because the `tests` package is not being added to the Python path by pytest and so the relative import doesn't work?

I think we can avoid having `tests` be a package by either wrapping the `gen_package` function into a fixture in `conftest.py`

```Python
import pytest

def _gen_package(
    path: pathlib.Path, project_config: dict[str, str]
) -> subprocess.CompletedProcess[str]:
    ...

@pytest.fixture
def gen_package():
    return _gen_package
```
and having the tests which use it accept a `gen_package` argument or refactoring it directly in to a fixture with `project_config` and `path`  also fixtures which [can then be overridden within a particular test module / test](https://docs.pytest.org/en/stable/how-to/fixtures.html#overriding-fixtures-on-various-levels)
```Python
import pathlib
import subprocess

import pytest

@pytest.fixture
def project_config():
    return {
        "github_username": "test-user",
        "project_short_description": "description",
        "project_name": "Cookiecutter Test",
        "expected_repo_name": "cookiecutter-test",
        "expected_package_name": "cookiecutter_test",
    }

@pytest.fixture
def package_path(tmp_path):
    return tmp_path


@pytest.fixture
def gen_package(
    package_path: pathlib.Path, project_config: dict[str, str]
) -> subprocess.CompletedProcess[str]:
    ...
```
There are also some other alternatives listed in the [answers to this StackOverflow question](https://stackoverflow.com/q/33508060).

tests/helpers.py
matt-graham: I'd personally prefer `generate_package` as I think abbreviations in variable / function names make it more difficult to remember which particular shortening is being used and also less explicit to someone who might not understand what `gen` means.

tests/test_git_init.py
dstansby: Agree it's not ideal, but I think this is how you have to specify parmeterized tests (see [pytest docs on this](https://docs.pytest.org/en/stable/example/parametrize.html))? Happy to take suggestions for improvements though.

tests/helpers.py
dstansby: I think from the principle of Keep It Simple, just a function is better here?

dstansby:
Thanks for comments! I won't have time to fix this in the next couple of weeks, so someone else feel free to take over if you're keen to get this in quicker!

tests/test_git_init.py
paddyroddy: I think, as a minimum, they don't need to be strings
```suggestion
    [True, False],
```
so can also change `if "True" in git_init_cookiecutter_option` to `if git_init_cookiecutter_option`

tests/helpers.py
paddyroddy: I really don't see the harm in `fixtures`

tests/test_git_init.py
samcunliffe: You can pass in a one-element dictionary and expand it when you define the config dictionary...

```python
@pytest.mark.parametrize(
    "one_element_dictionary",
    [{"initialise_git_repo": False}, {"initialise_git_repo": True}],
)
def test_append_to_dictionary(one_element_dictionary):
    config = {
        "github_username": "test-user",
        **one_element_dictionary,
    }

```

samcunliffe: Obviously you don't want to call it `one_element_dictionary`...

samcunliffe: You can pass in a one-element dictionary and expand it when you define the config dictionary...

```python
@pytest.mark.parametrize(
    "one_element_dictionary",
    [{"initialise_git_repo": False}, {"initialise_git_repo": True}],
)
def test_append_to_dictionary(one_element_dictionary):
    config = {
        "github_username": "test-user",
        **one_element_dictionary,
    }

```

tests/helpers.py
samcunliffe: Thinking again, do we _want_ this to be a fixture to get the tempdir teardown?

samcunliffe: But I tend to agree with David in spirit: not having decorators is fewer lines of code. And fewer lines is fewer lines.

tests/test_git_init.py
samcunliffe: Since I'm the one taking over, I did this. @paddyroddy - is [017b8f4](https://github.com/UCL-ARC/python-tooling/pull/405/commits/017b8f4b6d0d92ac55b15de684db5b2bfd23c570) more readable?

samcunliffe:
****

> At the moment the approach of making the `tests` directory a package by adding an `__init__.py` (presumably so we can use a relative import for `helpers` module) seems to break the ability to run tests directly by running `pytest` or `pytest -s` from the root of the repository (with the latter being what we currently recommend in the developer notes in the README), though tests do still run fine using `tox`. Specifically I get the following errors on test collection
> 
> ```
> _______________________________________ ERROR collecting tests/test_git_init.py ________________________________________
> ImportError while importing test module '.../python-tooling/tests/test_git_init.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> tests/test_git_init.py:8: in <module>
>     from .helpers import gen_package
> E   ModuleNotFoundError: No module named 'tests'
> ______________________________________ ERROR collecting tests/test_package_gen.py ______________________________________
> ImportError while importing test module '.../python-tooling/tests/test_package_gen.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> tests/test_package_gen.py:6: in <module>
>     from .helpers import gen_package
> E   ModuleNotFoundError: No module named 'tests'
> ```
> 
> I guess this is because the `tests` package is not being added to the Python path by pytest and so the relative import doesn't work?
> 
> I think we can avoid having `tests` be a package by either wrapping the `gen_package` function into a fixture in `conftest.py`
> 
> ```python
> import pytest
> 
> def _gen_package(
>     path: pathlib.Path, project_config: dict[str, str]
> ) -> subprocess.CompletedProcess[str]:
>     ...
> 
> @pytest.fixture
> def gen_package():
>     return _gen_package
> ```
> 
> and having the tests which use it accept a `gen_package` argument [...]

I went for your first suggestion. Adding fixtures as defaults to functions returned by fixtures ended up a bit messy. This way is explicit-er and only repeats one line in the two tests... so not too terrible?

tests/test_git_init.py
```diff
@@ -59,46 +41,46 @@ def test_initialisation_of_git_repo(
         text=True,
     )
 
-    if "True" in git_init_cookiecutter_option:
-        # git status should succeed
-        assert git_status.returncode == 0
```
samcunliffe, line 64: I just inverted this logic.

tests/test_git_init.py
paddyroddy: Yes sir

tests/conftest.py
```diff
@@ -1,24 +1,39 @@
...
-    -------
-    dict
-        dictionary with values for the cookiecutter template,
+    Arguments:
+    ---------
+    path: Path
+        Directory to create pacakge in.
+    project_config: dict
+        A dictionary with values for the cookiecutter template,
         as defined in the cookiecutter.json
```
samcunliffe:
```suggestion
    config: dict
        A dictionary with values for the cookiecutter template,
        as defined in the cookiecutter.json
    path: Path
        Directory to create package in.
```

tests/test_package_generation.py
```diff
@@ -0,0 +1,61 @@
+"""Checks that the cookiecutter works."""
+
+import pathlib
+import subprocess
+from typing import Callable
```
paddyroddy:
```suggestion
import typing
```
then use `typing.Callable`. @p-j-smith persuaded me on this school of thought, so it's easy to see where something came from (i.e. `pathlib.Path`)

tests/test_package_generation.py
```diff
@@ -0,0 +1,61 @@
...
+        "pyproject.toml",
+        "src",
+        pathlib.Path("src") / "cookiecutter_test",
+        pathlib.Path("src") / "cookiecutter_test" / "__init__.py",
+        "tests",
+        pathlib.Path(".github"),
+        pathlib.Path(".github") / "workflows",
+        "mkdocs.yml",
+        pathlib.Path("docs") / "index.md",
+        pathlib.Path("docs") / "api.md",
```
paddyroddy, line 38: Is this up-to-date?

tests/test_git_init.py
```diff
@@ -2,49 +2,31 @@
 
 import pathlib
 import subprocess
+from typing import Callable
```
paddyroddy: Same `import typing`

tests/test_git_init.py
```diff
@@ -2,49 +2,31 @@
...
-            git_init_cookiecutter_option,
-        ],
-        capture_output=True,
-        check=True,
-        text=True,
-    )
+    result = generate_package(config=test_config, path=tmp_path)
 
-    test_project_dir = tmp_path / project_config["expected_repo_name"]
+    test_project_dir = tmp_path / "cookiecutter-test"
```
paddyroddy, line 26: Would it be better to have that in `test_config`? Feels a bit hard-codey

tests/conftest.py
```diff
@@ -1,24 +1,39 @@
-"""Defines fixtures for tests."""
+"""Fixtures for the cookiecutter template tests."""
+
+import pathlib
+import subprocess
+from typing import Callable
```
paddyroddy: same `import typing`

paddyroddy (commented):
Same thoughts

tests/test_package_generation.py
samcunliffe: For you, I will do this.

tests/test_package_generation.py
samcunliffe: To my knowledge yes....?

tests/test_git_init.py
samcunliffe: Oh yeah. David changed it. Hmm... I don't really mind. We no longer have the "expected" keys in the config dictionary, because it's really just the input. Perhaps this way is clearer?

dstansby: The intention was to leave `test_config` as a dictionary exclusively for the cookiecutter config, hence removing any expected values from it.

tests/test_package_generation.py
dstansby: Presumably (hopefully!) the test would fail if it wasn't up to date

tests/test_git_init.py
```diff
@@ -2,49 +2,31 @@
...
+import typing
 
 import pytest
 
 
 @pytest.mark.parametrize(
-    "git_init_cookiecutter_option",
-    ["initialise_git_repository=True", "initialise_git_repository=False"],
+    "git_init",
+    [{"initialise_git_repository": True}, {"initialise_git_repository": False}],
```
dstansby: I think having two dictionaries for a single parametrized argument is more complicated/messier than just doing a simple parametrize like I was doing before.

In addition to this way being more complex, `pytest` won't generate nice test names. Before it would have been `test_initialisation_of_git_repo[git_init=True]`, but now it would be `test_initialisation_of_git_repo[git_init[{"initialise_git_repository": True}]`.

tests/test_git_init.py
```diff
@@ -2,49 +2,31 @@
...
-        git_init_cookiecutter_option: A string defined in cookiecutter.json,
-            which determines if git should be initialised or not
-
-    """
+    """Checks to see if git was correctly initialised if desired."""
+    test_config = {
+        "github_username": "test-user",
+        "project_short_description": "description",
+        "project_name": "Cookiecutter Test",
+        **git_init,
```
dstansby: And what you're improving above in explicitly naming the variable in the decorator, you're losing here by doing a 'blind' dictionary expansion which doesn't show the key/value pairs being passed into the `test_config` dict.

dstansby (commented):
Thanks for finishing! Can't approve my own PR, but LGTM apart from the suggestion of reverting back to a simpler pytest parametrization, but I don't think it's worth spending much time bikeshedding that.

tests/test_git_init.py
samcunliffe: 🤔 That is true. 

And the output of the tests in the logs will be read more often than the code. So I now tend to agree. 

It's really just one `git revert` able commit. So I'm fine with going back.

samcunliffe: I let Matt and Paddy express an opinion.

tests/test_package_generation.py
paddyroddy: It's for @p-j-smith more than me! But thank you

tests/test_package_generation.py
paddyroddy: Good point!

tests/test_git_init.py
paddyroddy: What about my `True`/`False` rather than `"True"`/`"False"` suggestion?

dstansby: Does ☝️ mean you are happy with reverting this change, but then "True"/"False" should still be changed to bools instead of strings?

paddyroddy: Yes

paddyroddy: (as long as it works, which I believe it should)

samcunliffe: It does... [fde687b](https://github.com/UCL-ARC/python-tooling/pull/405/commits/fde687b7fb127d1bf58cae33f1d0ca21246ec6d6).

samcunliffe: Incidentally, comments about readability are always welcome.
Bikes and sheds notwithstanding.

paddyroddy (approved):
🫐

Link checker failing, but assume that is unrelated

matt-graham (approved):
Thanks @samcunliffe for the updates and @dstansby for the initial work, and sorry for the delay re-reviewing. This all looks good to me now!

PR #406, opened 2024-06-07
Make funder optional

### LINKED ISSUE #290: [TASK]: Make funder option optional
### What needs to be done?

Currently, you are forced to pick a funder, this may not always be appropriate,

dstansby:
Fixes https://github.com/UCL-ARC/python-tooling/issues/290; depends on https://github.com/UCL-ARC/python-tooling/pull/405.

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -12,7 +12,7 @@
   "min_python_version": ["3.10", "3.11", "3.12"],
   "max_python_version": ["3.12", "3.11", "3.10"],
   "license": ["MIT", "BSD-3", "GPL-3.0"],
-  "funder": "JBFC: The Joe Bloggs Funding Council",
+  "funder": "",
   "__prompts__": {
     "author_given_names": "Given name(s) of package author",
     "author_family_names": "Family name(s) of package author",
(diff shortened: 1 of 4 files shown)


samcunliffe (approved):
This also looks great. 🫐 

I appreciated in today's workshop how we don't want to force this on users.

samcunliffe:
Is prettier complaining about the `{% if cookiecutter.funder != '' %}` clause? Because we're using `{% if...` elsewhere in template README.

dstansby:
Might be because without a funder the README was getting two newlines at the end? Lets see if my fix works...

paddyroddy:
So I've had a look and `prettier` currently fails if there's a funder or not.

Without funder - removes trailing line
image

With funder - two different lines removed
image

Let me have a play

paddyroddy:
I've fixed the no funder case

paddyroddy:
This is fixed now @dstansby. Basically, it was just some `jinja2` syntax missing. In general, I've found it best to have things in the following format (note the `-`):

```jinja2
{%- if something %}
some code
{%- endif %}
```

dstansby:
What does the `-` do?

paddyroddy (approved):
🫐

paddyroddy:
> What does the `-` do?

Just whitespace control https://www.redpill-linpro.com/techblog/2023/07/19/jinja_whitespaces.html

Issue #329, opened 2024-03-25
Add some regression tests on generated package files


dstansby:
### Is Your Feature Request Related to a Problem? Please Describe

We don't currently have any regression tests for the files that are generated by the package template. As an example of where this would have been useful, see https://github.com/UCL-ARC/python-tooling/pull/309. I made a mistake that broke a couple of the `pyproject.toml` fields, that would have been more easily spotted if there was a test that compared `pyproject.toml` against a reference/expected file.

### Describe the Solution You'd Like

Add a test that compares `pyproject.toml` to an expected/reference file, which fails if the generated file doesn't match. I would suggest starting with just `pyprojec.toml` to pin down the testing machinery, and then rolling out to other files in a follow up PR.

PR #409, opened 2024-06-07
Renaming `github_username` to `github_owner` + generating `__repo_name` & `__repo_url`

matt-graham:
Though its not exposed to user now we have longer prompts rather than using variable names directly, I think `github_username` is a bit of a misnomer given that it can be either a user or organization name, hence I think `github_owner` would be a better choice.

We also currently manually construct the GitHub repository URL in lots of different bits of the template and also the qualified repository name in a few places. We can use [double underscore prefixed (rendered) private variables](https://cookiecutter.readthedocs.io/en/stable/advanced/private_variables.html) in the cookiecutter config to generate the repository URL and qualified name once and then reuse elsewhere to avoid the repetition. This would also make it simpler to later switch to supporting alternative repository hosting options such as GitLab in future (which is the rational for naming the variables `__repo_name` and `__repo_url` rather than `__github_repo_name` and `__github_repo_url`).


dstansby:
I'm 👍 , but reluctant to approve before we have some regression tests (https://github.com/UCL-ARC/python-tooling/issues/329) since there's a lot of changes here that could be susceptible to human error (e.g., missing or adding "/" characters from URLs)

paddyroddy (approved):
Oo I didn't know about the private variables, big fan of this approach.

Are there any other private variables that we could leverage?

samcunliffe:
> I'm 👍 , but reluctant to approve before we have some regression tests (https://github.com/UCL-ARC/python-tooling/issues/329) since there's a lot of changes here that could be susceptible to human error (e.g., missing or adding "/" characters from URLs)

Isn't it enough the tests we currently have, all pass? We do do a cookiecut operation in CI, right? Which is working...

https://github.com/UCL-ARC/python-tooling/blob/2477eee282b58b8083e96013cae35843e10cf6b9/tests/test_package_gen.py#L7

samcunliffe (approved):
🫐 

(Conflicts, obvs)

matt-graham:
Tests were previously failing since #432 was merged in as the passed configs now need to use `github_owner` rather than `github_username` - now fixed and have also updated tutorial to make more consistent with new `github_owner` variable name.

samcunliffe (approved):
🫐

samcunliffe:
Let's make `v1.0.0` of the template when this is in.

paddyroddy (approved):
🫐

cookiecutter.json
```diff
@@ -8,11 +8,13 @@
   "project_short_description": "A cookiecutter package with UCL ARC recommendations.",
   "initialise_git_repository": true,
   "deploy_docs_to_github_pages": true,
-  "github_username": "{{cookiecutter.author_given_names.lower().replace(' ', '-')}}-{{cookiecutter.author_family_names.lower().replace(' ', '-')}}",
+  "github_owner": "{{cookiecutter.author_given_names.lower().replace(' ', '-')}}-{{cookiecutter.author_family_names.lower().replace(' ', '-')}}",
   "min_python_version": ["3.11", "3.12", "3.13"],
   "max_python_version": ["3.13", "3.12", "3.11"],
   "license": ["MIT", "BSD-3", "GPL-3.0"],
   "funder": "",
+  "__repo_name": "{{cookiecutter.github_owner}}/{{cookiecutter.project_slug}}",
```
Saransh-cpp, line 16: TIL `__variable` in `cookiecutter.json`, looks so clean!

PR #410, opened 2024-06-07
Default to enabling deploying docs to GitHub pages in template

### LINKED ISSUE #408: Default to enabling deploying docs to GitHub pages in template

With #404 now merged in the user gets relatively clear instructions on next steps needed to get deployment of HTML documentation to GitHub Pages working if `deploy_docs_to_github_pages` option is set to `True`, so I would argue it would make sense to now have this option set to `True` by default, as if a user does not have a strong preference otherwise (for example wanting to use another hosting solution like ReadTheDocs) I think defaulting to encouraging having documentation hosted somewhere would be good. However I realise this might be contentious so wanting to get opinions on this - @paddyroddy, @samcunliffe, @dstansby.

Changing default for cookiecutter option `deploy_docs_to_github_pages` to `True`. 

Leaving default as is.

### Additional Context

_No response_

matt-graham:
Would resolve #408

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -7,7 +7,7 @@
   "package_name": "{{cookiecutter.project_slug.replace('-', '_')}}",
   "project_short_description": "A cookiecutter package with UCL ARC recommendations.",
   "initialise_git_repository": true,
-  "deploy_docs_to_github_pages": false,
+  "deploy_docs_to_github_pages": true,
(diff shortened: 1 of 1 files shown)


paddyroddy (approved):
🫐

matt-graham:
Seems like there is a consensus in favour of this from #410 (or at least no objections so far) so merging!

PR #412, opened 2024-06-11
Fix `gh repo create` command for the Windows command prompt

samcunliffe:
In the *Festival of Digital Research and Scholarship* workshop: @llapira spotted that our command is broken on the Windows command prompt: needs double quotes otherwise it thinks this is five individual arguments.


samcunliffe:
@all-contributors please add @llapira for bug

paddyroddy (approved):
Good catch @LLapira!

PR #430, opened 2024-07-22
Improve expected file test 

dstansby:
As a first step towards https://github.com/UCL-ARC/python-tooling/issues/329, this improves the testing of the generated package files by generating a list of all files/folders in the generated package, and comparing that to an expected list.

This is an improvement over the previous test, where we only checked that several explicit files/folders exist. The new test will now catch cases where we add extra files/folders that aren't expected.

### DIFF
--- a/tests/test_package_generation.py
+++ b/tests/test_package_generation.py
@@ -84,7 +84,7 @@ def test_package_generation(
         a for a in actual_files if len(a.parts) > 1 and a.parts[0] == ".git"
     }
 
-    assert sorted(actual_files) == sorted(expected_files)
+    assert actual_files == expected_files
 
     # Check it's pip-installable
     pipinstall = subprocess.run(  # noqa: S603
--- a/tests/test_package_generation.py
+++ b/tests/test_package_generation.py
@@ -1,5 +1,6 @@
 """Checks that the cookiecutter works."""
 
+import os
 import pathlib
 import subprocess
 import typing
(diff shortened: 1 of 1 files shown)


tests/test_package_generation.py
```diff
@@ -24,26 +44,47 @@ def test_package_generation(
...
+        pathlib.Path("tests/test_dummy.py"),
+    }
+
+    actual_files = get_all_files_folders(test_project_dir)
+    # Filter out anything under .git/ to make comparison easier
+    actual_files = actual_files - {
+        a for a in actual_files if len(a.parts) > 1 and a.parts[0] == ".git"
+    }
+
+    assert sorted(actual_files) == sorted(expected_files)
```
samcunliffe: Does a set comparison need to be sorted?

samcunliffe (approved):
🫐

tests/test_package_generation.py
samcunliffe: ... No.

```
In [1]: lhs = set([1,2,3,4,5])

In [2]: rhs = set([3,2,4,5,1])

In [3]: lhs == rhs
Out[3]: True
```

samcunliffe:
```suggestion
    assert actual_files == expected_files
```

dstansby: It doesn't, but I did this so it printed in a deterministic order when it fails.

dstansby: Maybe it's a bit less confusing if it doesn't use sets at all

dstansby: Actually went to change this and think your suggestion is better 😄

PR #432, opened 2024-07-24
Add regression test for generated package data

### LINKED ISSUE #329: Add some regression tests on generated package files

dstansby:
This adds a regression test for generated pacakge data. The test folder now has a data folder inside it which contains a reference copy of the generated package. This is then compared at test time to a newly generated package. Any differences are shown after the failing test with a diff that looks like:

```
E           RuntimeError: Non-zero diff between generated files and expected files.
E           Generated files can be found in /Users/dstansby/software/python-tooling/pytest_output/test_package_generation0/cookiecutter-test.
E
E           --- /Users/dstansby/software/python-tooling/pytest_output/test_package_generation0/cookiecutter-test/README.md
E           +++ /Users/dstansby/software/python-tooling/tests/data/test_package_generation/README.md
E           @@ -119,7 +119,7 @@
E            tox -e docs
E            ```
E            
E           -from the root of the repository. The built docs will be written to
E           +from the root of the repository. The built documentation will be written to
E            `site`.
E            
E            Alternatively to build and preview the documentation locally, in a Python

tests/test_package_generation.py:79: RuntimeError
```

To update the test files with an intentional change, when the test fails the test data is updated with the changes that can then be easily comitted (if desired).

Fixes https://github.com/UCL-ARC/python-tooling/issues/329


pyproject.toml
```diff
@@ -1,13 +1,12 @@
...
-addopts = [
-    "--color=yes",
-    "--import-mode=importlib",
-    "--verbose",
-]
-testpaths = [
-    "tests",
+[tool.mypy]
+exclude = [
+    'tests\/data\/test_package_generation\/src\/cookiecutter_test\/__init__\.py',
```
dstansby: Can someone work out how I can get this to work? mypy doesn't seem to be ignoring the file 😢

paddyroddy: So I encountered this problem yesterday (along with @alessandrofelder) https://github.com/python/mypy/issues/17977. It's not ideal, but maybe we just exclude from `pre-commit`?

paddyroddy:

paddyroddy (changes requested):

pyproject.toml
```diff
@@ -1,13 +1,12 @@
...
 ]
 
+[tool.pytest.ini_options]
+addopts = ["--color=yes", "--import-mode=importlib", "--verbose"]
+testpaths = ["tests"]
```
paddyroddy: This should be in a separate PR

pyproject.toml
```diff
@@ -27,9 +26,7 @@ lint.per-file-ignores = {"*tests*" = [
     "INP001",
     "T201",
 ]}
-lint.select = [
-    "ALL",
-]
+lint.select = ["ALL"]
```
paddyroddy: Why are you changing this? It was consistent throughout the file

pyproject.toml
paddyroddy: I see what you've done now. This change shouldn't be done here. It's not in keeping with the rest of the file

dstansby:
> Are we sure we want this? Every time there's a new change, this test will have to be updated.

See https://github.com/UCL-ARC/python-tooling/issues/329 for context and motivation.

Definitely appreciate that this adds a bit more work when changing anything in the template, but I think it's worth it to catch potential errors such as in https://github.com/UCL-ARC/python-tooling/pull/309 which slipped through the PR author and reviewer.

I've tried to make the loop here as simple as possible:

1. Run tests
2. If this test fails, it modifies the test data in place
3. Review if the test data is modified as expected. If it is, commit and push the changes.

pyproject.toml
dstansby: Sorry, I had an overzealous linter. Should be fixed now

pyproject.toml
dstansby: Sorry, I had an overzealous linter. Should be fixed now

paddyroddy:
> See #329 for context and motivation.
> 
> Definitely appreciate that this adds a bit more work when changing anything in the template, but I think it's worth it to catch potential errors such as in #309 which slipped through the PR author and reviewer.

Consider me won over, I'd forgotten about #309

dstansby:
Sigh, putting this back as draft until I can fix the test

paddyroddy (changes requested):
Can you fix (by which I mean exclude) the link checking failures?
image

paddyroddy:
The tests are failing again @dstansby. `main` had updated.

paddyroddy (approved):
I've fixed this @dstansby. Will approve and merge now so it doesn't need fixing again.

PR #433, opened 2024-07-24
Make git init test more robust

dstansby:
This makes the git initialisation test more robust by explicitly specifying the directory that should be a git repository. Prevents a false positive if the cookiecutter package dir doesn't have a `.git` directory, but one its parent directories does.


samcunliffe:
This has been open for a while, sorry about that.

Would we want to support the case that a user had a parent directory repo? I think probably not, but just checking.

paddyroddy (approved):
Looks good

PR #448, opened 2024-09-27
No longer require the second question to be mandatory

paddyroddy:
I find this one a bit annoying sometimes


paddyroddy:
Binder down

samcunliffe (approved):

samcunliffe:
> Binder down

It's been 505 for a while now. I wonder why.

... Perhaps this might be a reason to demote it? [It's currently a 🟢](https://github.com/UCL-ARC/python-tooling/blob/main/docs/pages/libraries/jupyter-notebooks.md#live-executable-environments).

paddyroddy:
> It's been 505 for a while now. I wonder why.

Looks like it's back up https://mybinder.org

PR #460, opened 2024-10-11
Native TOML support for `tox`

### LINKED ISSUE #455: Adopt native TOML `tox` support

Following https://github.com/tox-dev/tox/issues/3353 https://github.com/ymyzk/tox-gh-actions/discussions/197, it will at some be possible to have full native TOML support. I tried this for now and couldn't get it to work, so we might need to wait a bit. This would involve removing the lecacy INI format.

_No response_

_No response_

### Additional Context

_No response_

paddyroddy:
Fixes #455. In order to do this, have also switched from `tox-gh-actions` to `tox-gh`. Has been tested in https://github.com/astro-informatics/sleplet/pull/420, https://github.com/astro-informatics/sleplet/pull/421.

- [x] Will need to check if templating works correctly

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,9 +40,9 @@ jobs:
           cache-dependency-path: pyproject.toml
 
       - name: Install dependencies
-        run: python -m pip install tox tox-gh-actions
+        run: python -m pip install tox tox-gh
 
       - name: Run tests
-        run: tox
+        run: tox run
         env:
-          OS: ${{ matrix.os }}
+          TOX_GH_MAJOR_MINOR: ${{ matrix.python-version }}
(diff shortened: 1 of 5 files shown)


dstansby:
I would just filter `.pyc` files out of the comparison (like the `.git` folder is) to make it a bit easier

paddyroddy:
> I would just filter `.pyc` files out of the comparison (like the `.git` folder is) to make it a bit easier

Ah, thanks, that bit was annoying me. Guessing it is a difference between `tox-gh` and `tox-gh-actions`.

paddyroddy:
Thanks @dstansby!

samcunliffe (approved):
This looks great. We should probably add it to our recommendations page? Because `toml`-er is better.

paddyroddy:
> This looks great. We should probably add it to our recommendations page? Because `toml`-er is better.

👍 have raised https://github.com/UCL-ARC/python-tooling/issues/461

PR #470, opened 2024-10-22
Add release workflow

paddyroddy:
This will create a `latest` tag that will track the `release` tag, so that we can encourage users to use the `latest` tag, but when submitting a bug report we could use the actual version.

Note, if you don't follow SemVer then this wouldn't work. But that is kind of unavoidable.

Have tested on a dummy repo.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,28 @@
+---
+name: Update Tag
+
+on:
+  release:
+    types:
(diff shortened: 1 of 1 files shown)


samcunliffe (approved):
LGTM!

(For reference: @paddyroddy showed me his test repo and we had a play.)

dstansby:
Ooh neat idea!

PR #472, opened 2024-10-22
Change to `github.ref_name` for `latest` flag

paddyroddy:
I've changed `github.sha` to `github.ref_name`.

- `github.sha` points to the current commit SHA in the workflow
- `github.ref_name` will contain the tag name of the release

This ensures the latest tag points to the same commit as the release tag, rather than potentially pointing to a different commit that triggered the workflow. This is especially important if you create releases through the GitHub UI, where you might select an older commit to create the release from.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -24,5 +24,5 @@ jobs:
         run: |-
           git config user.name github-actions
           git config user.email github-actions@github.com
-          git tag --force ${{ env.TAG_NAME }} ${{ github.sha }}
+          git tag --force ${{ env.TAG_NAME }} ${{ github.ref_name }}
           git push origin ${{ env.TAG_NAME }} --force


samcunliffe (approved):
Fine for me, but I thought SHA would work...

samcunliffe:
Should we mention the label in the CONTRIBUTING? I'd say probably not... just for the sake getting bogged down.

paddyroddy:
> Fine for me, but I thought SHA would work...

It does, but is `ref_name` safer?

paddyroddy:
> Should we mention the label in the CONTRIBUTING? I'd say probably not... just for the sake getting bogged down.

Yeah I reckon not worth it

matt-graham:
> This ensures the latest tag points to the same commit as the release tag, rather than potentially pointing to a different commit that triggered the workflow. This is especially important if you create releases through the GitHub UI, where you might select an older commit to create the release from.

I think in GitHub's terminology 'commit that triggered the workflow' has a meaning which is dependent on the triggering event and [for the `release` event](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release) `GITHUB_SHA` (and I assume therefore `github.sha`) is always the last commit in the tagged release. So I don't know that there is any risk of these being different at the point the workflow is run? But I guess there is a possibility the tag could be later changed to point to a different commit SHA in which case maybe pointing to the release tag is slightly preferable, though I would say we would probably not want to ever change _release_ tags after the fact.

paddyroddy:
> though I would say we would probably not want to ever change _release_ tags after the fact.

Agreed

matt-graham (approved):
Approving as doesn't seem like there is any downside to this change!

Issue #471, opened 2024-10-22
Once #470 is merged (and we're happy) we should make a `v1.0.0` release


paddyroddy:
### What needs to be done?

We should make a `v1.0.0` release. Then we should change the docs to say to use `cookiecutter https://github.com/UCL-ARC/python-tooling --checkout latest`. Then follow the steps documented in #260.

Issue #260, opened 2024-01-23
Versioning the template.


samcunliffe:
@matt-graham says we will get the latest commit at the HEAD of main unless specified.
- [x] verify this

Make a latest tag.

paddyroddy:
I have a workflow for tracking tags https://github.com/paddyroddy/.github/blob/main/.github/workflows/tags.yml

paddyroddy:
Things to do:
- [ ] Make a release (Blocked by #16)
- [x] Add a tracking `latest` tag, which auto updates
- [x] Document `--checkout latest` every we refer to it
- [x] Update bug template #issue #263

samcunliffe:
# Notes from hackmorning 2024-10-22

## Discussion

We discussed options and looked at how [MIRSG have set things up](https://github.com/UCL-MIRSG/.github/blob/2c76880ceb86f4575a103cf66b285c929c364891/.github/workflows/tags.yaml#L42). Their infrastructure is a bit more fancy: they have an auto-updating tag `v0` that always points to the newest `v0.x`. 

We don't need a major version following tag, we want a `latest` tag that follows the latest version (major or minor).

Sam's sketch of what we want (probably doesn't actually work as written...)
```yml
on:
    release:
        types:
            - created

jobs:
    update-tag:
        runs-on: ubuntu-latest
        steps:
            - name: Checkout source
              uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4

            - name: Set up git
              run: git config --global user.email "ACTIONS BOT"

            - name: Update "latest" tag to point to the newest release
              run: git tag -f latest $GITHUB_SHA && git push --tags
```

## Summary

Auto-updating `latest` tag pointing to the SHA of the newest release tag.

## Decision

@paddyroddy will create a short GitHub action that runs on a release, takes the SHA of the release tag and runs a `git tag -f latest` to (re)create and push a `"latest"` tag.  Then we can have instructions in `README.md` that point to latest.

paddyroddy:
We now have:
- #476 
- #475

dstansby:
What's the motivation to add manual releases, instead of the versioning just being the SHA of whichever commit you have checked out?

As far as I can see adding manual releases makes things quite a lot complicated, for little gain. But maybe there's something I'm missing?

paddyroddy:
[...]

PR #476, opened 2024-10-22
Once #471 is merged, we should encorage checkout `latest`

paddyroddy:
Part of #260

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ and how to use the newly created package with some of the tools included.
 2. Run cookiecutter in the desired directory
 
    ```sh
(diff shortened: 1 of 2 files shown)


samcunliffe (approved):
Oh! There's nothing stopping us tagging the `HEAD` of `main` now (perhaps on merge of this).

Then _these_ instructions will work, even before we finalise #471.

paddyroddy:
> Oh! There's nothing stopping us tagging the `HEAD` of `main` now (perhaps on merge of this).
> 
> Then _these_ instructions will work, even before we finalise #471.

Good point! Will merge this and make the tag

Issue #463, opened 2024-10-17
Revisit using pre-commit.ci


dstansby:
If I'm makng a simple change to markdown, such as in https://github.com/UCL-ARC/python-tooling/pull/462, I use the GitHub web interface for speed. This can often result in pre-commit failing however, because the markdown isn't correctly linted. This then breaks the benefit of using the GitHub website for speed, because I have no way of automatically fixing the linting and have to go back to local development on my laptop.

To improve this I'd like to propose we revisit using pre-commit.ci to lint this repo. This would provide a mechanism of automatically making pre-commit fixes on demand, by posting a "pre-commit.ci autofix" comment in a PR.

In https://github.com/UCL-ARC/python-tooling/pull/196 we very briefly discussed and decided not to use it because it's not available for private repositories. One mitigation to this is we can still have a GitHub actions linting job, so when someone sets up a private repo from this template there's still a linting CI job that is run.

So I'd like to advocate for using pre-commit.ci for this repo. Thoughts?

samcunliffe:
> In https://github.com/UCL-ARC/python-tooling/pull/196 we very briefly discussed and decided not to use it because it's not available for private repositories. One mitigation to this is we can still have a GitHub actions linting job, so when someone sets up a private repo from this template there's still a linting CI job that is run.

Briefly from memory the decision to turn it off was:

1. We didn't want to have different tools in the main tooling repo than those in the template.
2. pre-commit.ci is not free for closed-source: we want to encourage maximum takeup of the template amongst UCL researchers, and a good portion of projects either start out closed until some research outcome is published, or stay closed-source.

I'm actually OK with violating 1. We could have pre-commit.ci on the repo, but not the template.

If I read you correctly, you're suggesting that we make the template a bit smarter and have it choose whether to lint based on a `cookiecutter` option?

paddyroddy:
I'm happy to have it on the repo, but not the template

dstansby:
I'm definitely only proposing for the repo here, not the template.

I'm not suggesting we alter the template - just that we add `pre-commit.ci` to this repo. In order to make sure we keep [eating our own dog food](https://en.wikipedia.org/wiki/Eating_your_own_dog_food), my suggestion is we could keep the GitHub actions linting running on this repo too. But that would duplicate linting CI on this repo 🤔 I'm not sure which way round is best...

paddyroddy:
[...]

PR #482, opened 2024-10-29
Revert "[pre-commit.ci] pre-commit autoupdate"

paddyroddy:
Reverts UCL-ARC/python-tooling#479

We don't want the alpha version


paddyroddy:
Another issue is this isn't going to update the template's ones. We don't want this on at all.

dstansby (dismissed):
I think it's good to keep these up to date, so see no reason to revert this. If you want to disable pre-commit.ci auto updating the versions, please could you open a new issue and we can discuss there?

paddyroddy:
@renovatebot will do it in 3 days...

samcunliffe:
🤷 

I'm ok with alphas. But if we decide never to take pre-commit autoupdates and only use @renovate: can we actually disable it in the `ci:` config, please? [It's not immediately clear to me but I guess it's possible](https://pre-commit.ci/#configuration).

dstansby:
I've re-opened https://github.com/UCL-ARC/python-tooling/issues/463 to discuss, so we can make sure discussion is in one place for future reference

paddyroddy:
> I'm ok with alphas. But if we decide never to take pre-commit autoupdates and only use @renovate: can we actually disable it in the `ci:` config, please? [It's not immediately clear to me but I guess it's possible](https://pre-commit.ci/#configuration).

It's not possible https://github.com/pre-commit-ci/issues/issues/83

dstansby:
@paddyroddy could you post the issue you've found on this thread? I think they're spread around several PRs/issues at the moment, and I'm struggling to find them all 😄

paddyroddy:
Also the problem is this doesn't update the template which @renovatebot would

samcunliffe (approved):
Let's revert this and add the bot back?
- https://github.com/UCL-ARC/python-tooling/issues/463#issuecomment-2451500351

PR #502, opened 2024-12-03
Force test `pre-commit` config to be picked up

### LINKED ISSUE #503: @renovatebot bumping versions will break the regression test every time

paddyroddy:
Fixes #503. I want to fix this properly upstream, but this is a hack that we've used in MIRSG https://github.com/UCL-MIRSG/.github/blob/e61d159d27e610d165272c8ad5f1b743dabc4ff0/.renovaterc.json5#L7-L9. I think the regex that @renovatebot uses only looks for one folder deep, i.e. https://docs.renovatebot.com/modules/manager/pre-commit/#default-config.

### DIFF
--- a/.renovaterc.json5
+++ b/.renovaterc.json5
@@ -5,4 +5,7 @@
     ":assignAndReview(team:collaborations-python-tooling)",
     ":maintainLockFilesMonthly",
   ],
+  "pre-commit": {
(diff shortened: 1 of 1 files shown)


paddyroddy:
Replaced/improved with #505.

PR #504, opened 2024-12-03
Fix #503 by excluding version numbers from the regression tests.

### LINKED ISSUE #503: @renovatebot bumping versions will break the regression test every time

samcunliffe:


samcunliffe:
Pipped to the post by #505.

PR #513, opened 2025-02-02
Persist light theme on refresh 

### LINKED ISSUE #414: Light mode doesn't persist on navigation/refresh
### Describe the Bug

When browsing the website in light mode, the theme switches on every refresh. 

### To Reproduce

Switch to light mode and refresh the website. 

### Expected Behaviour

The website stays in light mode. 

### Actual Behaviour

The website switches theme back to dark mode. 

### Additional Context

```Markdown
- Operating system: Tested in Chrome Android/mobile and Firefox on Mac.
```

samcunliffe:
Fixes 

- #414

... using the workaround suggested in 

- https://github.com/just-the-docs/just-the-docs/issues/1223#issuecomment-1509704703.

Worth spending a bit of thought on whether we want more lines of my hacky JS code saving the theme to local storage, which we potentially can't delete until [jtd/#1223](https://github.com/just-the-docs/just-the-docs/issues/1223) is solved nicely upstream.

### Question:

Is it worth switching build system to something that supports theme toggling and inferring theme from the browser/os settings? It feels like it's the responsibility of the doc engine/doc theme. E.g. [Material for MkDocs supports it](https://github.com/UCL-ARC/mkdocs-rc-docs/blob/140e8fe5eef0379997af1b2eaaf396786bc006b8/mkdocs-project-dir/mkdocs.yml#L28).

That said, if you're happy with these lines, I am.

### DIFF
--- a/docs/_includes/header_custom.html
+++ b/docs/_includes/header_custom.html
@@ -2,9 +2,22 @@
 
 <script>
   const toggleDarkMode = document.querySelector(".js-toggle-dark-mode");
-
   jtd.addEvent(toggleDarkMode, "click", function () {
     if (jtd.getTheme() === "light") {
+      jtd.setTheme("dark");
+      toggleDarkMode.textContent = "☼";
+      localStorage.setItem("theme", "dark");
+    } else {
+      jtd.setTheme("light");
+      toggleDarkMode.textContent = "☾";
+      localStorage.setItem("theme", "light");
+    }
+  });
+
+  // TODO: we can probably delete this (hacky workaround) if/when just-the-docs/#1223 is fixed.
+  // Meanwhile, we check each time if there is a theme written to local storage and obey it.
+  window.addEventListener("DOMContentLoaded", function () {
+    if (localStorage.getItem("theme") === "dark") {
       jtd.setTheme("dark");
       toggleDarkMode.textContent = "☼";
     } else {
--- a/docs/_includes/header_custom.html
+++ b/docs/_includes/header_custom.html
@@ -14,8 +14,9 @@
     }
   });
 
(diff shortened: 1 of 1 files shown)


matt-graham (commented):
I think on balance having a few lines extra JS for a more functional site in the interim is worth the tradeoff, but agree it seems that using an alternative solution for building the site like Material for MkDocs which support this out of the box would be good.

When serving this locally and viewing in Firefox the theme still does not seem to persist between page changes, which from adding some `console.log` statements seems to be as the `"DOMContentLoaded"` event listener never gets called. As far as I can tell (my JavaScript knowledge is very minimal though!) the syntax for adding the event listener is all correct, so I am not sure if this is just something specific about serving this locally?

Saransh-cpp (commented):
+1 on does not work for me locally - both on refreshing and changing pages

Migrating to another docs generator sounds like the best option here

samcunliffe:
Ooh you're right! 

[Perhaps we're hitting this](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete)? It was working yesterday on the train, but I now can reproduce your problem.

samcunliffe:
Now working for me...

https://github.com/user-attachments/assets/47026c5e-b2f3-4760-99f0-319da87c958b

matt-graham (approved):
Not sure why it did, but with change to comment this does now work for me locally!

samcunliffe:
> Not sure why it did, but with change to comment this does now work for me locally!

Python is cool. Javascript is not cool.

Saransh-cpp (approved):
Nice! Works well now!

paddyroddy:
➕ for moving to https://squidfunk.github.io/mkdocs-material as well

Issue #516, opened 2025-02-10
`test_pip_installable` test make persistent changes to environment test is run in


matt-graham:
### Describe the Bug

The test in `tests/test_package_generation.py::test_pip_installable` 

https://github.com/UCL-ARC/python-tooling/blob/a3003e2be9b1278a1d1965039f3f4eef89dc0e86/tests/test_package_generation.py#L95-L123

installs an instance of the template with the package name `cookiecutter-test` in the Python environment the test is run in, and this installation is not removed after test completion (successful or not).

While the actual generated package directory is created in a temporary directory that is cleared, the references to this package in the environment `site-packages` directory are not cleaned up.

I think we probably should create a virtual environment in a temporary directory and tests the package installation there. While we could just remove the package after checking installation, this has the downside that if we in future add any default dependencies to the package template then we would need to remember to also remove these.

### To Reproduce

Create a clean virtual environment, install `pytest`. Run `pip list` to see initial installed package

Run

```
pytest tests/test_package_generation.py::test_pip_installable
```

from root of repository. 

Run `pip list` again and compare to previous output. 

### Expected Behaviour

No persistent changes to environment are made by tests, in particular no hanging package installation references are left - specifically `pip list` output is same before and after running tests.

### Actual Behaviour

`pip list` output changed after running tests - there will be an additional `cookiecutter-test` package listed pointing to a now non-existent temporary directory.

### Version In Use

v1.0.0-31-ga3003e2

### Additional Context

```Markdown
- Cookiecutter version: 2.6.0
- Operating system: Ubuntu
- Python version: 3.13.0
```

PR #517, opened 2025-02-10
Fix MkDocs build from template instances

### LINKED ISSUE #515: MkDocs build from template broken
### Describe the Bug

While using template to create https://github.com/UCL/causalprog I came across a few issues with building MkDocs site

- The TOML `tox` specification of the `docs` environment uses the wrong syntax for specifying the commands to run - there should be an additional set of square brackets as commands is expected to be an array of arrays, each inner array corresponding to a single command (potentially consisting of multiple components).
- Our `mkdocs.yml` configuration in the template is out of date with current MkDocs - specifically, inventories should now be specified with `inventories` rather than `imports` key, and `docstring_style` should live under `handlers.python.options`

Changes required to template are equivalent to those in
https://github.com/UCL/causalprog/compare/0e4d06c..130dbaa

### To Reproduce

Create an instance of template and run `tox -e docs` or `mkdocs build`

### Expected Behaviour

No errors when attempting to build documentation site locally or on GitHub Actions

### Actual Behaviour

A series of errors (first one from `tox` TOML config problem, latter two from `mkdocs.yml` issues

`TypeError: 'mkdocs' is not list`

`TypeError: PythonConfig.__init__() got an unexpected keyword argument 'docstring_style'`

```
INFO    -  FutureWarning: The 'import' key is renamed 'inventories' for the Python handler
             File "causalprog/.tox/docs/lib/python3.13/site-packages/mkdocstrings/handlers/base.py", line 680, in get_handler
               self._handlers[name] = module.get_handler(**kwargs)
             File "causalprog/.tox/docs/lib/python3.13/site-packages/mkdocstrings_handlers/python/handler.py", line 361, in get_handler
               warn("The 'import' key is renamed 'inventories' for the Python handler", FutureWarning, stacklevel=1)
```

### Version In Use

v1.0.0-31-ga3003e2

### Additional Context

```Markdown
- Cookiecutter version: 2.6.0
- Operating system: Ubuntu
- Python version: 3.13.0
```

matt-graham:
Resolves #515 

- Updates `mkdocs.yml` configuration in template to use structure required by current MkDocs.
- Fixes TOML `tox` configuration for `docs` environment to use required array of arrays syntax.

Leaving this as a draft for now as we probably want to add a test for building MkDocs site from a template instance to catch things like this in future, ideally by running `tox -e docs` as well as `mkdocs build` to also catch error in `tox` configuration. I am not sure whether we would want to add `tox` as a development dependency here, or create a temporary virtual environment as suggested also in #516, install `tox` there and test running `tox -e docs` (plus potentially default test environment)?

### DIFF
--- a/tests/test_package_generation.py
+++ b/tests/test_package_generation.py
@@ -129,6 +129,7 @@ def test_optional_funder(
     tmp_path: pathlib.Path, generate_package: typing.Callable, funder: str
 ) -> None:
     """Test package generation."""
+    """Test specifying funder or not in package generation."""
     config = {
         "github_owner": "test-user",
         "project_short_description": "description",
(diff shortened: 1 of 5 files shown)


matt-graham:
If we do go with including `pytest-venv` as a test dependency in #519 then that would be an easy way to also test running `mkdocs` build by installing `mkdocs` and `tox` in a temporary venv.

tests/test_package_generation.py
```diff
@@ -129,6 +129,7 @@ def test_optional_funder(
     tmp_path: pathlib.Path, generate_package: typing.Callable, funder: str
 ) -> None:
     """Test package generation."""
+    """Test specifying funder or not in package generation."""
```
paddyroddy: I assume you didn't mean to have two docstrings?

matt-graham: Nope! Now fixed thanks

paddyroddy (approved):
Looks good to me

PR #518, opened 2025-02-11
docs: update tutorial to use uv

### LINKED ISSUE #477: Update the tutorial to not use `conda` (switch to `uv`?)
Notably when working on 
- #464 

I _did not_ change `tutorial.md`. I took a look and realised that I needed to test any `uv` instructions on a Windows computer (since, presumably many of our users will be Windows). An advantage of Anaconda is that you get a terminal with it.

Saransh-cpp:
Removed conda and introduced uv for everything.
Reviewed the tutorial on a Windows machine.

Closes #477


tutorial.md
```diff
@@ -9,49 +9,30 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
 
 <details><summary>Click to expand... </summary> <!-- markdownlint-disable-line MD033 -->
 
-To use the template you will need to have at least the following software tools installed,
+To start using the template you will need at least the following software tools installed:
```
Saransh-cpp:
```suggestion
To use the template you will need at least the following software tools installed
```

tutorial.md
```diff
@@ -9,49 +9,30 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
+- [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
+- [uv](https://docs.astral.sh/uv/getting-started/installation/)
 
-and ideally also,
+Additionally, [the GitHub command line interface (CLI)](https://cli.github.com/), [pre-commit](https://pre-commit.com/), and [tox](https://tox.wiki) are required to complete some of the follow on steps for using the package generated with the template, so while they are not strictly needed, you will get more out of the tutorial if you have them installed. For the follow on exercises you will also need an [account set up on GitHub](https://github.com/join) if you don't already have one.
 
-- [the GitHub command line interface (CLI)](https://cli.github.com/),
-- [pre-commit](https://pre-commit.com/),
-- [tox](https://tox.wiki).
+An easy way to install the software you need is by using `uv`, which can create a dedicated environment with the necessary packages installed using the step by step instructions below:
```
Saransh-cpp:
```suggestion
An easy way to install the software you need is by using `uv`, which can create a dedicated environment with the necessary packages installed using the step by step instructions below
```

tutorial.md
```diff
@@ -9,49 +9,30 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
-
-1. Download and install [Miniconda](https://docs.anaconda.com/free/miniconda/) following the operating system specific instructions at either
-
-   - <https://docs.anaconda.com/free/miniconda/miniconda-install> for a graphical installer
-   - or <https://docs.anaconda.com/free/miniconda/#quick-command-line-install> for installation via the command line.
-
-   If you already have Miniconda (or Anaconda) installed you can skip this step.
-
-2. Once Miniconda is installed, you need to open a terminal window:
+1. Open a terminal window:
```
Saransh-cpp:
```suggestion
1. Open a terminal window
```

tutorial.md
```diff
@@ -9,49 +9,30 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
 
-   - On Windows: open the Start menu from the taskbar, type `miniconda` in the search field, then click `Anaconda Prompt (miniconda3)` from the results (or `Anaconda Prompt (anaconda3)` if using a previous Anaconda installation).
+   - On Windows: open the Start menu from the taskbar, type `cmd` in the search field, then click [command prompt](https://en.wikipedia.org/wiki/Command-line_interface#Command_prompt) from the results.
    - On MacOS: click the Launchpad icon in the Dock, type `Terminal` in the search field, then click `Terminal` from the results.
    - On Linux: open the default terminal application installed in your distribution.
 
-3. Once you have a terminal window open you should see text `(base)` as part of your [command prompt](https://en.wikipedia.org/wiki/Command-line_interface#Command_prompt) if Minconda has been installed and set up correctly, indicating you currently have the default `base` environment active. If this is the case you should run
+2. In the opened terminal window, use:
```
Saransh-cpp:
```suggestion
2. In the terminal window, use
```

tutorial.md
```diff
@@ -9,49 +9,30 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
+   to install the "tools" necessary for creating and using a package using the `UCL-ARC/python-tooling` cookiecutter template.
 
-4. To check that all the dependencies have installed correctly in the new environment first activate the environment by running
-
-   ```sh
-   conda activate python-tooling
-   ```
-
-   at which point you should see `(base)` in the command prompt change to `(python-tooling)`, Then try running each of the following commands in turn, one at a time,
+3. To check that all the dependencies have installed correctly, run each of the following commands in turn, one at a time,
```
Saransh-cpp:
```suggestion
3. To check that all the dependencies have installed correctly, run each of the following commands in turn, one at a time
```

tutorial.md
```diff
@@ -90,30 +71,21 @@ An easy way to get all the software you need installed is using the tool [Conda]
...
 
-1. Open a terminal window:
-
-   - On Windows: open the Start menu from the taskbar, type `miniconda` in the search field, then click `Anaconda Prompt (miniconda3)` from the results.
-   - On MacOS: click the Launchpad icon in the Dock, type `Terminal` in the search field, then click `Terminal` from the results.
-   - On Linux: open the default terminal application installed in your distribution.
-
-2. In the opened terminal window change the working directory to the path you wish to create the package in using the `cd` (change directory) command.
-3. Activate the `python-tooling` Conda environment you previously created ([see instructions above](#%EF%B8%8F-setting-up-dependencies-for-using-template)) by running
+1. Open a new terminal/command prompt window and change the working directory to the path you wish to create the package in using the `cd` (change directory) command.
```
Saransh-cpp:
```suggestion
1. Open a new terminal/command prompt window and change the working directory to the path you wish to create the package in using the `cd` (change directory) command
```

tutorial.md
```diff
@@ -9,49 +9,33 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
    - On Linux: open the default terminal application installed in your distribution.
 
-3. Once you have a terminal window open you should see text `(base)` as part of your [command prompt](https://en.wikipedia.org/wiki/Command-line_interface#Command_prompt) if Minconda has been installed and set up correctly, indicating you currently have the default `base` environment active. If this is the case you should run
+2. In the terminal window, use
 
    ```sh
-   conda create -y -n python-tooling -c conda-forge cookiecutter git gh pre-commit tox
+   uv tool install cookiecutter
+   uv tool install pre-commit
+   uv tool install tox
```
samcunliffe: Do they really need `tox` and `pre-commit`?

samcunliffe (commented):
Thanks so much for picking this up!

I'm probably not a good reviewer though: I only have access to macOS and Ubuntu. 

(Kinda related: should we think about running the tutorial again in the next ARC Festival of ...? I guess that's more of a question to the ether.)

tutorial.md
```diff
@@ -9,49 +9,33 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
+To use the template you will need at least the following software tools installed
 
-- [cookiecutter](https://cookiecutter.readthedocs.io/en/stable/),
-- [Git](https://git-scm.com/),
+- [Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
+- [uv](https://docs.astral.sh/uv/getting-started/installation/)
+- [the GitHub command line interface (CLI)](https://github.com/cli/cli?tab=readme-ov-file#installation)
 
-and ideally also,
+Additionally, [pre-commit](https://pre-commit.com/) and [tox](https://tox.wiki) are required to complete some of the follow on steps for using the package generated with the template, so while they are not strictly needed, you will get more out of the tutorial if you have them installed. For the follow on exercises you will also need an [account set up on GitHub](https://github.com/join) if you don't already have one.
```
paddyroddy:
```suggestion
Additionally, [pre-commit](https://pre-commit.com/) and [tox](https://tox.wiki) are required to complete some of the follow along steps for using the package generated with the template, so while they are not strictly needed, you will get more out of the tutorial if you have them installed. For the follow along exercises you will also need an [account set up on GitHub](https://github.com/join) if you don't already have one.
```

tutorial.md
```diff
@@ -9,49 +9,33 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
-
-2. Once Miniconda is installed, you need to open a terminal window:
-
-   - On Windows: open the Start menu from the taskbar, type `miniconda` in the search field, then click `Anaconda Prompt (miniconda3)` from the results (or `Anaconda Prompt (anaconda3)` if using a previous Anaconda installation).
+   - On Windows: open the Start menu from the taskbar, type `cmd` in the search field, then click [command prompt](https://en.wikipedia.org/wiki/Command-line_interface#Command_prompt) from the results.
    - On MacOS: click the Launchpad icon in the Dock, type `Terminal` in the search field, then click `Terminal` from the results.
    - On Linux: open the default terminal application installed in your distribution.
 
-3. Once you have a terminal window open you should see text `(base)` as part of your [command prompt](https://en.wikipedia.org/wiki/Command-line_interface#Command_prompt) if Minconda has been installed and set up correctly, indicating you currently have the default `base` environment active. If this is the case you should run
+2. In the terminal window, use
```
paddyroddy:
```suggestion
2. In the terminal window, type
```

tutorial.md
paddyroddy: Can these be one line?

tutorial.md
```diff
@@ -223,25 +198,47 @@ The index page of the documentation reproduces the content from the repository R
...
+
+```sh
+uv venv --python 3.11.6
+```
 
-An alternative to Conda is the [`venv` module](https://docs.python.org/3/library/venv.html) built-in to the Python standard library. This has the advantage of being available in any Python (3.3+) environment, but unlike Conda will not by itself allow you to use a different Python version from the system level install. In contrast to Conda which by default creates all environments in a shared user-level directory (if using Miniconda, by default in a directory `miniconda3/envs` in your user or home directory), the `venv` module requires being passed a path in which to create the directory containing the files associated with the virtual environment. A common pattern is to store the virtual environment files in a directory `.venv` within the root directory of the project repository. This can be achieved by running
+Alternatively, you can use the [`venv` module](https://docs.python.org/3/library/venv.html), which is slower, has less features, and is not developed rapidly when compared to `uv`, but is built-in to the Python standard library. Similar to `uv`, a virtual environment can be created using
```
paddyroddy:
```suggestion
Alternatively, you can use the [`venv` module](https://docs.python.org/3/library/venv.html), which is slower, has fewer features, and is not developed rapidly when compared to `uv`, but is built-in to the Python standard library. Similar to `uv`, a virtual environment can be created using
```

paddyroddy (approved):
A few comments

tutorial.md
matt-graham: ["Follow on" is correct here for the sense of  "things following on from"](https://dictionary.cambridge.org/dictionary/english/follow-on) which is what I intended when I originally wrote this in the context of running this as a tutorial at the festival. "Follow along" would I think change the meaning

tutorial.md
matt-graham: Not with one `uv` command - `uv tool install --help` gives usage as

```
Usage: uv tool install [OPTIONS] <PACKAGE>
```

and trying with multiple package names gives an error.

It might be possible to stack multiple commands on a line using separators but I'm not sure if this would be easy to do in a way that will work across operating system / shells and also would probably be less readable.

tutorial.md
```diff
@@ -9,49 +9,33 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
+- [uv](https://docs.astral.sh/uv/getting-started/installation/)
+- [the GitHub command line interface (CLI)](https://github.com/cli/cli?tab=readme-ov-file#installation)
 
-and ideally also,
+Additionally, [pre-commit](https://pre-commit.com/) and [tox](https://tox.wiki) are required to complete some of the follow on steps for using the package generated with the template, so while they are not strictly needed, you will get more out of the tutorial if you have them installed. For the follow on exercises you will also need an [account set up on GitHub](https://github.com/join) if you don't already have one.
 
-- [the GitHub command line interface (CLI)](https://cli.github.com/),
-- [pre-commit](https://pre-commit.com/),
-- [tox](https://tox.wiki).
+An easy way to install the software you need is by using `uv`, which can create a dedicated environment with the necessary packages installed using the step by step instructions below
```
matt-graham: While I think `uv` can be used to handle the Python based tools needed (`cookiecutter`, `tox`, `pre-commit`) it doesn't provide a solution for the GitHub CLI or Git itself which would need to be separately installed (while `conda` does allow installing non Python dependencies). I don't think this is a blocker to recommending using `uv` but I think this section will need to be reworked slightly to indicate Git and GitHub CLI will need to be installed separately and with links to relevant installation instructions.

tutorial.md
matt-graham: I'd say that the specific point made previously that the built in `venv` module does not allow you to manage the base Python version used in the environment would be useful to keep here. I would also say that 'developed rapidly' is not necessarily always a benefit 🙃 particularly for infrequent / less experienced users for which a more stable tool might be better.

matt-graham (commented):
Thanks @Saransh-cpp for the updates, this mostly looks great to me. I've tested setting up `uv` on Windows and installing the tools as explained in the text and it all worked fine. I've added a few comments - main one is that I think the section on installing dependencies may need reworking a bit to make it clearer that Git and GitHub CLI will need to be installed separately, as previously `conda` was also being used to install these.

tutorial.md
paddyroddy: When reading it, I genuinely wasn't sure what was intended. If it's that use case then I think we should use following.

tutorial.md
Saransh-cpp: We don't! Thanks!

Saransh-cpp: We don't need `pre-commit` and `tox` so that should solve it.

tutorial.md
Saransh-cpp: Added a sentence ✅

tutorial.md
Saransh-cpp: Should be back now

tutorial.md
```diff
@@ -9,67 +9,43 @@ In this tutorial we will go through in detail the steps required to set-up a Pyt
...
+- [uv](https://docs.astral.sh/uv/getting-started/installation/)
+- [the GitHub command line interface (CLI)](https://github.com/cli/cli?tab=readme-ov-file#installation)
 
-and ideally also,
+The instructions for installing these tools can be accessed by clicking the individual links above.
 
-- [the GitHub command line interface (CLI)](https://cli.github.com/),
-- [pre-commit](https://pre-commit.com/),
-- [tox](https://tox.wiki).
+Additionally, [cookiecutter](https://cookiecutter.readthedocs.io) will be required to generate your package from them template, which can be installed using the instructions below. For the follow on exercises you will also need an [account set up on GitHub](https://github.com/join) if you don't already have one.
```
Saransh-cpp:
```suggestion
Additionally, [cookiecutter](https://cookiecutter.readthedocs.io) will be required to generate your package from the template, which can be installed using the instructions below. For the follow on exercises you will also need an [account set up on GitHub](https://github.com/join) if you don't already have one.
```

matt-graham (approved):
Thanks @Saransh-cpp for the updates, all looks good to me 😍!

tutorial.md
```diff
@@ -223,25 +196,47 @@ The index page of the documentation reproduces the content from the repository R
...
+Alternatively, you can use the [`venv` module](https://docs.python.org/3/library/venv.html), which is slower and has fewer features, when compared to `uv`, but is built-in to the Python standard library. `venv` has the advantage of being available in any Python (3.3+) environment, but unlike `uv` will not by itself allow you to use a different Python version from the system level install. A common pattern is to store the virtual environment files in a directory `.venv` within the root directory of the project repository. This can be achieved by running
```
samcunliffe, line 249: I just ran through these for sanity as a fresh user in macOS.

The only stumbling point is from here on, it's not clear that we should preferentially be using UV.

That Is, if I skip or skim read the "Alternatively..." part and then try to `python -m pip install X` I will get:
```
(python-template) teaching@scafell python-template % python -m pip install numpy
/Users/teaching/python-template/.venv/bin/python: No module named pip
```
Obviously because I didn't install `pip` in my `venv`. But I _should_ be running `uv pip install` all the time.

Should we cull some of the information that follows? Or put it into a `<details></details>` block to hide it away?

I think we want to end the `</details>` block around
```
263 Once you have created and activated a virtual environment for the project, ...
```
Because the `uv` users (who will be in the majority?) will also need to install their own package into the `uv venv`.

samcunliffe (commented):
Aaah down on 289 we're telling them to use `tox`. So perhaps we should tell them to install `tox` around there.

```
uv tool install tox
```
... ? Ditto `pre-commit` down at 312. 
(Sorry @Saransh-cpp I gave you a wrong comment 🙃 )

But perhaps installing them should come later around the relevant section of the tutorial?

Saransh-cpp:
> Aaah also down on 289 we're ...

Line 266 teaches them to install the dev and test dependencies for the package, and `tox` + `pre-commit` are bundled within those groups (which should actually use `uv`, fixing this too) 😉 

I think this is getting shadowed by the whole `venv` stuff, which yes, should be in a `</details>` block. I'll update it. Thanks!

samcunliffe:
> Line 266 teaches them to install the dev and test dependencies for the package, and `tox` + `pre-commit` are bundled within those groups (which should actually use `uv`, fixing this too) 😉

Oh, great!

samcunliffe (approved):
Perfect. Cheers!

samcunliffe:
I forgot!

@all-contributors please add @Saransh-cpp for docs, content, maintenance, review

PR #519, opened 2025-02-13
Test package installation in temporary virtual environment

### LINKED ISSUE #516: `test_pip_installable` test make persistent changes to environment test is run in

matt-graham:
Resolves #516 

Creates a virtual environment in temporary package directory and tests package installation in there rather than installing in parent environment tests are run from.

Calling `pip install` command by directly referencing Python executable in created virtual environment seems to work in terms of ensuring package is installed in `site-packages` of temporary virtual environment rather than parent environment (at least I no longer see a `cookiecutter-test` entry when running `pip list` after running tests) and it seemed non trivial to `source` the virtual environment script with `subprocess.run`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,3 @@
-[[tool.mypy.overrides]]
-ignore_missing_imports = true
-module = ["pytest_venv"]
-
 [tool.pytest.ini_options]
 addopts = [
     "--color=yes",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[[tool.mypy.overrides]]
+ignore_missing_imports = true
+module = ["pytest_venv"]
+
 [tool.pytest.ini_options]
 addopts = [
     "--color=yes",
@@ -54,6 +58,7 @@ env_run_base = {commands = [
 ], deps = [
     "cookiecutter",
     "pytest",
+    "pytest-venv",
 ], description = "Test package creation", skip_install = true}
 gh.python = {"3.11" = [
     "py311",
--- a/tests/test_package_generation.py
+++ b/tests/test_package_generation.py
@@ -110,10 +110,15 @@ def test_pip_installable(
     subprocess.run(  # noqa: S603
         [sys.executable, "-m", "venv", str(venv_path)], check=True
     )
+    # Location of virtual environment Python executable is OS dependent
+    if sys.platform == "win32":
+        venv_python = str(venv_path / "Scripts" / "python.exe")
+    else:
+        venv_python = str(venv_path / "bin" / "python")
     # Try to install package in virtual environment with pip
     pipinstall = subprocess.run(  # noqa: S603
         [
-            str(venv_path / "bin" / "python"),
+            venv_python,
             "-m",
             "pip",
             "install",
--- a/tests/test_package_generation.py
+++ b/tests/test_package_generation.py
@@ -5,6 +5,7 @@ import os
 import pathlib
 import shutil
 import subprocess
+import sys
 import typing
 
 import pytest
@@ -103,12 +104,16 @@ def test_pip_installable(
         "project_name": "Cookiecutter Test",
     }
     generate_package(config=test_config, path=tmp_path)
-
-    # Check project directory exists
     test_project_dir = tmp_path / "cookiecutter-test"
+    venv_path = test_project_dir / ".venv"
+    # Create temporary virtual environment
+    subprocess.run(  # noqa: S603
+        [sys.executable, "-m", "venv", str(venv_path)], check=True
+    )
+    # Try to install package in virtual environment with pip
     pipinstall = subprocess.run(  # noqa: S603
-        [  # noqa: S607
-            "python",
+        [
+            str(venv_path / "bin" / "python"),
             "-m",
             "pip",
             "install",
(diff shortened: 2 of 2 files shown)


matt-graham:
I don't think the link checker check fail is anything to do with changes in this PR as it doesn't change or add any URLs. It looks like it might be due to
```
[304] https://discord.com/ | Redirect (304 Not Modified): Not Modified
```
flagging `https://discord.com/` as a redirect, potentially because of the trailing slash?

samcunliffe (commented):
Looks great!

2 ± 1 things come to mind:
* Did you look at [`pytest-venv`](https://github.com/mmerickel/pytest-venv)? It's unfortunately not receiving so much dev (last commit ~2 years ago). But it (or something similar) might cut down on some of the boilerplate?
* Alternatively we could plop all of this into a fixture (i.e. do as they do in the `pytest-venv` plugin but with our own code.
* The whole package generation test... um is there a smarter way to do it (I appreciate that is less of a comment for this bugfix and more of a general unhelpful musing).

matt-graham:
>  * Did you look at [`pytest-venv`](https://github.com/mmerickel/pytest-venv)? It's unfortunately not receiving so much dev (last commit ~2 years ago). But it (or something similar) might cut down on some of the boilerplate?
>  * Alternatively we could plop all of this into a fixture (i.e. do as they do in the `pytest-venv` plugin but with our own code.

I hadn't come across `pytest-venv` and my cursory searching for `pytest temporary virtual environment` didn't throw it up, but it does look like it does exactly what we want including dealing with the Windows specific directory layout that caught me out. While the lack of recent commits is a bit of a flag, it is a very minimal package so can imagine there isn't too much that can break and so need maintenance. I'd be happy to add it as a test dependency rather than reinventing the wheel ourselves. Any opinions to the contrary?

paddyroddy:
> The whole package generation test... um is there a smarter way to do it

There must be. I do think it's currently causing more problems than it has solved.

Saransh-cpp:
There is https://github.com/hackebrot/pytest-cookies, but it is unmaintained as well (tho used for testing in https://github.com/osl-incubator/scicookie).

If we end up migrating to copier, we can use https://github.com/12rambau/pytest-copie.

matt-graham:
Updated to use `pytest_venv` fixture rather than manual solution. `mypy` was complaining that
```
error: Cannot find implementation or library stub for module named "pytest_venv"  [import-not-found]
```
so I added a manual override to ignore import. Alternative would be not to add a type annotation for `venv` argument and add a `noqa` comment to stop Ruff complaining about this.

matt-graham:
Link checker flaking again (tried one re-run) this time due to
```
[503] https://coveralls.io/pricing | Network error: Service Unavailable
```

paddyroddy:
> Link checker flaking again (tried one re-run) this time due to
> 
> ```
> [503] https://coveralls.io/pricing | Network error: Service Unavailable
> ```

Coveralls is currently down

samcunliffe (approved):
Nice!

I'd say the new dependency is worth it for the one-liner environmentisation of our messy tests. (I think I've invented "environmentisation".)

tests/test_package_generation.py
```diff
@@ -8,6 +8,7 @@
 import typing
 
 import pytest
+import pytest_venv
```
paddyroddy: What about this one-liner instead?
```suggestion
import pytest_venv  # type: ignore[import-not-found]
```

pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[[tool.mypy.overrides]]
+ignore_missing_imports = true
+module = ["pytest_venv"]
```
paddyroddy: If we keep this, then I'd vote for splitting like this like the rest of the file. Means shorter diffs if we add to it and consistency.
```suggestion
module = [
	"pytest_venv",
]
```

paddyroddy (commented):
Nice, `pytest_venv` looks much neater than previous iteration

Saransh-cpp (approved):
Nice! 🚀

pyproject.toml
matt-graham: Have went with suggestion for inline ignore as seemed cleaner so marking this as resolved.

PR #528, opened 2025-04-02
Change `license.file` to `license-files` + SPDX-compliance

paddyroddy:
I encountered this issue when making a release https://github.com/astro-informatics/sleplet/pull/451. Also, make `BSD-3` SPDX-compliant.

### DIFF
--- a/tests/data/test_package_generation/pyproject.toml
+++ b/tests/data/test_package_generation/pyproject.toml
@@ -26,6 +26,7 @@ dynamic = [
 ]
 keywords = [
 ]
+license = "MIT"
 license-files = [
     "LICENSE.md",
 ]
--- a/tests/data/test_package_generation/pyproject.toml
+++ b/tests/data/test_package_generation/pyproject.toml
@@ -26,6 +26,9 @@ dynamic = [
 ]
 keywords = [
 ]
+license-files = [
+    "LICENSE.md",
+]
 name = "cookiecutter-test"
 optional-dependencies = {dev = [
     "build",
@@ -46,7 +49,6 @@ optional-dependencies = {dev = [
 ]}
 readme = "README.md"
 requires-python = ">=3.11"
-license.file = "LICENSE.md"
 urls.homepage = "https://github.com/test-user/cookiecutter-test"
 
 [tool.coverage]
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -29,6 +29,9 @@ dynamic = [
 ]
 keywords = [
 ]
+license-files = [
+    "LICENSE.md",
+]
 name = "{{cookiecutter.project_slug}}"
 optional-dependencies = {dev = [
     "build",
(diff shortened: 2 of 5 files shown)


matt-graham (approved):
Looks good to me. 

In case helpful to anyone understanding why this change is required, [relevant part of Python Packaging User Guide documentation](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files) says that

> As per [PEP 639](https://peps.python.org/pep-0639/) licenses should be declared with two fields:
>
> - `license` is an [SPDX license expression](https://packaging.python.org/en/latest/glossary/#term-License-Expression) consisting of one or more [license identifiers](https://packaging.python.org/en/latest/glossary/#term-License-Identifier).
> - `license-files` is a list of license file glob patterns.
>
> A previous PEP had specified license to be a table with a file or a text key, this format is now deprecated.

matt-graham (changes requested):
Actually reading the above linked documentation a bit more carefully I think we want to specify both `license` __and__ `license-files` fields?

{{cookiecutter.project_slug}}/pyproject.toml
```diff
@@ -29,6 +29,9 @@ dynamic = [
 ]
 keywords = [
 ]
+license-files = [
+    "LICENSE.md",
+]
```
matt-graham, line 35:
```suggestion
license = "{{cookiecutter.license}}"
license-files = [
    "LICENSE.md",
]
```

Assuming `cookiecutter.license` is always a valid [SPDX identifier](https://spdx.org/licenses/). Of the three current options `MIT` and `GPL-3.0` appear to be valid, but it might be that `BSD-3` should instead be `BSD-3-Clause` unless shortening is allowed?

tests/data/test_package_generation/pyproject.toml
```diff
@@ -26,6 +26,9 @@ dynamic = [
 ]
 keywords = [
 ]
+license-files = [
+    "LICENSE.md",
+]
```
matt-graham, line 32:
```suggestion
license = "MIT"
license-files = [
    "LICENSE.md",
]
```

matt-graham (changes requested):

{{cookiecutter.project_slug}}/pyproject.toml
paddyroddy: So from the [official list](https://spdx.org/licenses) it would suggest `BSD-3` is not valid.

From [the PEP](https://peps.python.org/pep-0639) I see no mention of shortening. So I have renamed to `BSD-3-Clause`.

paddyroddy:
https://stackoverflow.com is currently down

paddyroddy:
> Actually reading the above linked documentation a bit more carefully I think we want to specify both `license` **and** `license-files` fields?

I glanced over it initially, but yeah I think you are right, it does want both. It didn't error with just one, but did with the use of the old `license.file`.

samcunliffe (approved):

matt-graham (approved):
Thanks for updating, looks all good to me.

PR #541, opened 2025-06-02
Add Vale for content linting to the repo

paddyroddy:
In this PR I add [vale](https://vale.sh/) similarly to https://github.com/paddyroddy/talks/pull/80 https://github.com/paddyroddy/talks/pull/82. See this blog post for an explanation of what it is https://www.datadoghq.com/blog/engineering/how-we-use-vale-to-improve-our-documentation-editing-process. I used this extensively during my PhD https://github.com/paddyroddy/phd-thesis/blob/main/.vale.ini. I'm going to talk about this in the Collaboration Hour, so this PR is in prep for that.

The talk: https://paddyroddy.github.io/talks/linting-prose-with-vale.


.github/styles/custom/british-spellings.yml
samcunliffe, line 1: Is there an upstream place or option we can automagically use these from? Feels like it shouldn't be our responsibility to maintain a file of the correct spelling of 151 words.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -9,7 +9,7 @@ body:
     attributes:
       label: Is Your Feature Request Related to a Problem? Please Describe
       description: A clear and concise description of what the problem is.
-      placeholder: I'm always frustrated when [...]
+      placeholder: I'm always frustrated when […]
```
samcunliffe: ?

.github/styles/custom/endash.yml
samcunliffe, line 1: Ditto. I like the sentiment. But should this go in here?

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 # MIT License
 
-Copyright (c) 2023-2025 UCL Centre for Advanced Research Computing
+Copyright © 2023-2025 UCL Centre for Advanced Research Computing
```
samcunliffe: Oh. It's from `vale`. Probs not this one, right?

docs/pages/docs.md
```diff
@@ -5,7 +5,7 @@ layout: default
 
 ## Documentation
 
-With Python, as with many other languages, it's very common to automatically
+With Python, as with many other languages, it's common to automatically
```
samcunliffe: Cool simplification.

samcunliffe (commented):
So I really like the language simplifications. I think we want this 🫐 .

This was just a quick pass before seeing your talk. Curious about uncluttering the custom config. Could we make our own ARC `vale` language style erm... theme? (Not sure what's the correct word)

docs/LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022-2025 just-the-docs
+Copyright © 2022-2025 just-the-docs
```
samcunliffe: Probably good to blanket ignore the license files so we don't accidentally change the wording.

{{cookiecutter.project_slug}}/LICENSE.md
```diff
@@ -649,14 +649,14 @@ Also add information on how to contact you by electronic and paper mail.
...
-    <program>  Copyright (C) <year>  <name of author>
+    <program>  Copyright © <year>  <name of author>
     This program comes with ABSOLUTELY NO WARRANTY; for details type 'show w'.
     This is free software, and you are welcome to redistribute it
     under certain conditions; type 'show c' for details.
 
 The hypothetical commands `show w` and `show c` should show the appropriate
 parts of the General Public License. Of course, your program's commands might be
-different; for a GUI interface, you would use an “about box”.
+different; for a GUI, you would use an “about box”.
```
samcunliffe: Yeah this, for example, is a bad change. The GPL says you cant change the wording of the license.

samcunliffe (changes requested):
Sorry. Realised we need to ignore the license files in both the repo and the template.

tests/data/test_package_generation/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -21,7 +21,7 @@ body:
       value: |-
         import cookiecutter_test
 
-        ...
+        …
```
matt-graham: I'd say we want to disable the ellipsis replacement because of places like this - this is Python code in a YAML file and I suspect we'll have other cases like this where we have for example Python code in a Markdown file that we don't want it touch. While it doesn't really matter here as this is meant more to indicate user to fill in rest of reproducing code, while `...` is valid Python syntax `…` raises a syntax error ('Invalid character').

tutorial.md
```diff
@@ -80,7 +80,7 @@ We will first go through the steps for creating a new package using the `UCL-ARC
...
 
-3. Once you have completed all the cookiecutter prompts some additional instructions will be printed to screen (which we will come back to in the next sections) and your new package will be generated in a directory named `{project_slug}` in the current working directory (where `{project_slug}` is the value entered for the `'Slugified' project name...`[^slug] prompt, this will be `python-template` if you used the default placeholder values). You can see the directory tree of files generated by running
+3. Once you have completed all the cookiecutter prompts some additional instructions will be printed to screen (which we will come back to in the next sections) and your new package will be generated in a directory named `{project_slug}` in the current working directory (where `{project_slug}` is the value entered for the `'Slugified' project name…`[^slug] prompt, this will be `python-template` if you used the default placeholder values). You can see the directory tree of files generated by running
```
matt-graham: Think this ellipsis replacement we probably don't want as when rendered in fixed width font I think the Unicode character is less readable than the explicit `...`

{{cookiecutter.project_slug}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -21,7 +21,7 @@ body:
       value: |-
         import {{cookiecutter.package_name}}
 
-        ...
+        …
```
matt-graham: Same comment as above with regards to I think full `...` is better here as valid Python code

tutorial.md
```diff
@@ -129,7 +129,7 @@ GitHub CLI detected, you can create a repo with the following:
 gh repo create {github_owner}/{project_slug} -d "{project_description}" --public -r origin --source {project_slug}
 ```
 
-where `{github_owner}`, `{project_slug}` and `{project_description}` are replaced with the relevant GitHub repository owner (user name), project slug and description that you entered when setting up the package using `cookiecutter`. You should now copy-paste and run the `gh repo create ...` command that was printed out in the message - if you no longer have access to the message you can use the example above, being careful to subsitute the placeholders in braces `{}` with the relevant values.
+where `{github_owner}`, `{project_slug}` and `{project_description}` are replaced with the relevant GitHub repository owner (user name), project slug and description that you entered when setting up the package using `cookiecutter`. You should now copy-paste and run the `gh repo create …` command that was printed out in the message - if you no longer have access to the message you can use the example above, being careful to subsitute the placeholders in braces `{}` with the relevant values.
```
matt-graham: Again would say this ellipsis replacement decreases readability

.github/styles/custom/endash.yml
matt-graham: I agreed with @jeremyestein's comments after Collaborations hour talk that there are valid uses of en-dash &ndash; that don't correspnd to hyphenated words or as a replacement for parenthesis&mdash;for example separating names as in _Metropolis&ndash;Hastings algorithm_ or ranges _2020&ndash;2025_.

.github/styles/custom/british-spellings.yml
```diff
@@ -0,0 +1,151 @@
...
+  centers: centres
+  fibers: fibres
+  lusters: lustres
+  meters: metres
+  theaters: theatres
+  # -ise vs. -ize
+  '(\w+)ization\b': "$1isation"
+  '(\w+)ize\b': "$1ise"
+  '(\w+)ized\b': "$1ised"
+  '(\w+)izing\b': "$1ising"
```
matt-graham: I think there is a question of which convention we should go with to begin with here ([-ize and -ise forms are both accepted and in common use in British English](https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-ise,_-ize_(-isation,_-ization))) but more relevantly: will the current rules not lead to a lot of incorrect 'fixes'? For example _prize &rightarrow; prise_, _sizing &rightarrow; sising_, _seized &rightarrow; seised_

.github/styles/custom/british-spellings.yml
```diff
@@ -0,0 +1,151 @@
...
+  # add other common forms:
+  '(\w+)izable\b': "$1isable"
```
matt-graham: Also _sizable &rightarrow; sisable_

.github/styles/custom/no-contractions.yml
matt-graham, line 1: I'd be against excluding contractions. Often style guides recommend using contractions to enhance readability and increase conciseness (something I need all the help I can get with 😁) - for example the [gov.uk guide](https://www.gov.uk/guidance/content-design/writing-for-gov-uk).

.github/styles/custom/spaces.yml
matt-graham, line 1: This one seems reasonable but would want to make sure it doesn't for example affect Python code snippets included in Markdown documents where changing whitespace would break things.

README.md
```diff
@@ -15,11 +15,15 @@ repo, and we welcome questions there or in the `#helpme` channel on the
 🍪 Our template is a [cookiecutter] template which automatically creates new
 Python packages with our recommended tooling set up and ready to go.
 
+<!-- vale proselint.Annotations = NO -->
```
matt-graham: What was previously getting changed here that necessitated adding a manual exception / ignore block? The _you're_?

docs/pages/ci.md
```diff
@@ -27,7 +27,7 @@ during tests.
...
 | [Codecov](https://docs.codecov.com/docs) | Hosted service to report code coverage metrics. Occasionally slow to update after a report is updated, can be configured to add extra CI checks. This service is probably more widely used and is [free for both open-source and private projects](https://about.codecov.io/pricing/). | <span class="label label-green">Best</span> |
-| [Coveralls](https://docs.coveralls.io/)  | Hosted service to report code coverage metrics. Very similar to codecov and we don't strongly recommend one over the other. This service is only [free for open-source projects](https://coveralls.io/pricing).                                                                        | <span class="label label-green">Best</span> |
+| [Coveralls](https://docs.coveralls.io/)  | Hosted service to report code coverage metrics. Similar to codecov and we don't strongly recommend one over the other. This service is only [free for open-source projects](https://coveralls.io/pricing).                                                                             | <span class="label label-green">Best</span> |
```
matt-graham: Not sure I'm a fan of removing our agency in adding words like _very_ to emphasize statements.

matt-graham (changes requested):
I like the idea of having a prose linter but I would say we probably want some discussion of what rules to include (possibly this should be `needs-2-reviewers`?). I personally don't think the ellipsis replacement should be included as I would say the marginal gain in readability of the Unicode character in cases where it renders nicely (when not using a fixed-width font) are outweighed by the cases where it decreases readability (when showing in a fixed-width font for example) or might change the meaning of Python code snippets embedded in non-Python files. I also think the -ize/-ization/-ized/-izable to -ise/-isation/-ised/-isable replacements as currently implemented might end up with spurious matches, and I'd personally vote for not standardizing (😛) on -ise endings to begin with (but agree with at least having a consistent convention). I also find the removal of all _very_ instances a bit odd.

samcunliffe:
> I would say we probably want some discussion of what rules to include (possibly this should be needs-2-reviewers?)

🗳️ This guy wants: readability and accessibility suggestions inline in my PRs that have a lot of prose. I'm actually mostly interested in the accessibility of what I write. 

... Should we run the `Readability` rules and see what it suggests?

I'm not super bothered by consistency of contractions or em-dashes (though I understand the "-ize" sentiment, I would support consistent British spelling and fewer hedging and weasel words).

docs/pages/ci.md
samcunliffe: But we can ignore the suggestions, right?

Here "very" is needed.

.github/styles/custom/british-spellings.yml
samcunliffe: 🤣 That's a bug report to the original author @paddyroddy ?

paddyroddy:
> I like the idea of having a prose linter but I would say we probably want some discussion of what rules to include

This was very much for demo purposes. Obviously, we'd need to decide on what rules we want.

.github/styles/custom/british-spellings.yml
paddyroddy: Decided to remove this

.github/ISSUE_TEMPLATE/feature_request.yml
paddyroddy: https://www.compart.com/en/unicode/U+2026

.github/styles/custom/endash.yml
paddyroddy: This was used as an example. I hadn't refined the rule. Have turned it off...

LICENSE.md
paddyroddy: What is the practice involving licences? Once made can never change? Why do we use the `(c)` here?

docs/pages/docs.md
paddyroddy: I agree

docs/LICENSE
paddyroddy: Have pushed a way to do this

{{cookiecutter.project_slug}}/LICENSE.md
paddyroddy: https://github.com/UCL-ARC/python-tooling/pull/541#discussion_r2130307180

tests/data/test_package_generation/.github/ISSUE_TEMPLATE/bug_report.yml
paddyroddy: That is a fair enough reason. @K-Meech raised the point of screen readers for prose though.

paddyroddy: This file shouldn't have changed as not a Markdown file. This is me copy-and-pasting

tutorial.md
paddyroddy: https://github.com/UCL-ARC/python-tooling/pull/541#discussion_r2130314136

{{cookiecutter.project_slug}}/.github/ISSUE_TEMPLATE/bug_report.yml
paddyroddy: https://github.com/UCL-ARC/python-tooling/pull/541#discussion_r2130315506

tutorial.md
paddyroddy: https://github.com/UCL-ARC/python-tooling/pull/541#discussion_r2130314136

.github/styles/custom/british-spellings.yml
paddyroddy: https://github.com/UCL-ARC/python-tooling/pull/541#discussion_r2130287843

.github/styles/custom/british-spellings.yml
paddyroddy: It was obviously a rushed example

.github/styles/custom/no-contractions.yml
paddyroddy: It's a discussion point, or "it is"

.github/styles/custom/spaces.yml
paddyroddy: Have removed

paddyroddy: You can exclude code blocks FYI

README.md
paddyroddy: It was specifically the Markdown callout

docs/pages/ci.md
paddyroddy: Yes can ignore suggestions. I find the use of very triggering.

paddyroddy:
@samcunliffe @matt-graham I have reverted the changes, excluded licence files, made CI not error. To see the output you need to click on
image
we can then iterate on rules to turn on/off

Can be seen more clearly here
image

tests/data/test_package_generation/.github/ISSUE_TEMPLATE/bug_report.yml
K-Meech: @paddyroddy - I'm not sure what you mean by 'screen readers for prose' here? As far as I'm aware screen readers will handle both ellipsis styles similarly

.vale.ini
```diff
@@ -0,0 +1,24 @@
...
+write-good.Passive = NO
+write-good.TooWordy = NO
+
+# `vale sync` in CI means the packages are then linted, so turn them off
+[.github/styles/**]
+BasedOnStyles =
+
+# Don't modify licence files
+[**/LICENSE*]
+BasedOnStyles =
```
samcunliffe: Should we exclude the CODE_OF_CONDUCT.md too? It's not _legally_ problematic if we change the wording. But the contributor covenant is a pretty standard one. I'd say we don't want to accidentally change it.

And `write-good` is complaining about the weasel word "fairly" 🤣 

Screenshot 2025-06-06 at 09 27 40

samcunliffe (approved):

LICENSE.md
samcunliffe: Well the GPL specifically says "thou shalt not change this file". 

MIT doesn't have any such clause. But to avoid potential problems, I would not reword license files. 

`(c)` → `©` is probably innocent enough. I would leave as is, but ¯\\\_(ツ)_/¯

samcunliffe: Oh wait I realise you mean why the `(c)` at all: because it's not in the [MIT template](https://opensource.org/license/mit). That's a good point. Did it come from GitHub?

tests/data/test_package_generation/.github/ISSUE_TEMPLATE/bug_report.yml
paddyroddy: Sorry, I thought that was something said the other day. I have no idea what they do with ellipsis.

.vale.ini
paddyroddy: What do you think @matt-graham?

LICENSE.md
paddyroddy: I don't know where it came from, it is a bit weird though.

paddyroddy: I don't know enough about licences, but it does seem weird that I can't reword my own licence. Guess it makes sense for GPL. Like what if you made a typo?

.vale.ini
```diff
@@ -0,0 +1,27 @@
...
+                write-good
+
+# Disable
+Vale.Spelling = NO
+write-good.E-Prime = NO
+write-good.Passive = NO
+write-good.TooWordy = NO
+
+# Ignore lines starting with "> [!NOTE]"
+BlockIgnores = (?s)> \[!NOTE\].*?(\n|$)
```
paddyroddy: This replaces the inline comment before @matt-graham

samcunliffe: Eeeeeh what about `TIP` `IMPORTANT` `WARNING` `CAUTION`? Do they get flagged.
I could imagine `> [!TIP]` is quite useful.

```suggestion
BlockIgnores = (?s)> \[!(CAUTION|IMPORTANT|NOTE|TIP|WARNING)\].*?(\n|$)
```

paddyroddy: Yeah they would be

.vale.ini
matt-graham: I'd agree with Sam that as we're using an off the shelf code of conduct, better to keep it as-is and so exclude from linting.

.vale.ini
matt-graham: Ah so it was the `> ![NOTE]` alert syntax specifically that was triggering not the content of the alert? I guess we could also include the other alert types here (`!TIP`, `!IMPORTANT`, `!WARNING`, `!CAUTION`). I don't entirely understand the regex here - it looks like Value uses [`regexp2`](https://github.com/dlclark/regexp2) which is based on [Go `regexp`](https://pkg.go.dev/regexp/syntax), but after reading the docs I'm still entirely sure what `(?s)` bit is doing here (appears to be setting a flag to let `.` match `\n` but not sure why this is needed and whether it applies just to the group?). I think something like

```
^ *> \[!(NOTE|IMPORTANT|WARNING|CAUTION|TIP)\] *$
```
would avoid unintentionally capturing other characters with `.` while still allowing for white space before or after alert syntax (could also use `\w` in place of ` ` but not sure we actually want to allow for other white space types).

matt-graham: Just spotted that you've already covered all altert types in last commit 73fd5c1 - I still don't entirely understand the regex but it works ~~so marking as resolved~~! (EDIT: I hadn't refreshed before commenting so hadn't seem Sam's comment either).

.vale.ini
```diff
@@ -0,0 +1,27 @@
...
+                Vale,\
+                write-good
+
+# Disable
+Vale.Spelling = NO
+write-good.E-Prime = NO
+write-good.Passive = NO
+write-good.TooWordy = NO
+
+# Ignore lines starting with "> [!NOTE]"
```
matt-graham:
```suggestion
# Ignore lines starting with GitHub alerts syntax
```

matt-graham (approved):
Other than potentially excluding code of conduct looks good to me, thanks @paddyroddy!

.vale.ini
paddyroddy: I looked at the `regex101` guide here: https://vale.sh/docs/keys/blockignores. I've tested both regexes and they both work. But if you understand yours, I'll go with that.

paddyroddy: Also, I assume you meant `\s` in place of ` `? Hadn't occurred to me that `\s` was any type of whitespace. This is just in this flavour of regex? I've always taken it to mean strictly ` `.

paddyroddy: Ah bad news, yours doesn't work

matt-graham: So in the [example linked to in the docs](https://regex101.com/r/mFM0kZ/1/) there it looks like it's excluding the whole block `{< file }` delimited block. Are we trying to similarly ignore the context of callouts or just the syntax which starts an alert? From https://regex101.com/r/rvRsMy/1 it looks like we just match the initial bit not the content. If so I think the regex I suggested will do so as well but is tighter / less chance for false matches

paddyroddy: Went with `(?s)> \[!(CAUTION|IMPORTANT|NOTE|TIP|WARNING)\](\n|$)` and tested on `regex101`

paddyroddy: Yeah we want to just ignore the top line itself, not the contents within it. Your regex doesn't work, feel free to try it yourself locally (run `vale sync; vale .`)

Issue #537, opened 2025-05-23
Add a text label to the theme dark/light mode-changing button


samcunliffe:
See parent issue for explanation in James' email.

## Definition of done

The mode-changing button has an alt-text or meaningful label: It doesn't say "moon" (or "sun") when read by a screen reader. 

## Docs

Solution could be an [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label).

samcunliffe:
Huh. Should have been auto-closed. Probably I didn't use the correct word. 
Solved via #543.

paddyroddy:
> Huh. Should have been auto-closed. Probably I didn't use the correct word. Solved via [#543](https://github.com/UCL-ARC/python-tooling/pull/543).

You had Fixes
* #num

I think you have to do Fixes #num

PR #543, opened 2025-06-03
Add label to theme-switcher button.

samcunliffe:
Fixes 
- #537 

as suggested by @jamesprime by adding an [aria label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) to labelless button. Confirmed with macOS's "VoiceOver" screen reader...

Screenshot 2025-06-03 at 13 25 10

---

As with the ☼ / ☾ it's a bit WET because it needs declaring and then the same lines appear in the switch conditional.

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -13,15 +13,15 @@ GEM
       http_parser.rb (~> 0)
     eventmachine (1.2.7)
     ffi (1.17.2)
-    ffi (1.17.2-aarch64-linux)
+    ffi (1.17.2-aarch64-linux-gnu)
     ffi (1.17.2-aarch64-linux-musl)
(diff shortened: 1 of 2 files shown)


docs/Gemfile.lock
samcunliffe, line 1: 🙅

dstansby:
Could we upstream this into the theme source?

samcunliffe:
> Could we upstream this into the theme source?

Yep. Deleting this whole custom thing is waiting on https://github.com/just-the-docs/just-the-docs/issues/1223 (which I think is https://github.com/just-the-docs/just-the-docs/pull/1099)

paddyroddy (approved):
Lovely accessible work

PR #547, opened 2025-06-09
Stop `vale` failing for forked contributions that don't touch prose.

samcunliffe:
Relates to 
- #549

(i.e. the prose lint failure for pull requests from forked repositories).
Should improve the quality of life of external contributing devs.

This change _only_ annotates the diff introduced by a PR. So #549 will still be a bug for external contributors who write a lot of prose. But for technical contributions, at least, we avoid the 10 annotations limit.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -71,5 +71,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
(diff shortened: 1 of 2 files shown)


paddyroddy (approved):
Good non-bot.

Hopefully we can work out a _proper_ fix for this.

See explanations of the modes here https://github.com/reviewdog/reviewdog#filter-mode.

PR #555, opened 2025-07-15
Simplify install instructions

dstansby:
My 'hot' take is that the goal of the install instructions should be a single direct recommended way to use the template, with minimal diversions. Here I've changed the alternative install instructions to still link to the `cookiecutter` install instructions, but to shorten it to reduce potential diversion from the recommended way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ uv pip install -e ".[dev]"
 [website]: https://github-pages.arc.ucl.ac.uk/python-tooling
 [UCL ARC]: https://ucl.ac.uk/arc
 [cookiecutter]: https://cookiecutter.readthedocs.io/en/stable
+[install cookiecutter]: https://cookiecutter.readthedocs.io/en/stable/README.html#installation
 [uv]: https://docs.astral.sh/uv
 <!-- prettier-ignore-end -->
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,7 +66,6 @@ uv pip install -e ".[dev]"
 [website]: https://github-pages.arc.ucl.ac.uk/python-tooling
 [UCL ARC]: https://ucl.ac.uk/arc
 [cookiecutter]: https://cookiecutter.readthedocs.io/en/stable
-[install cookiecutter]: https://cookiecutter.readthedocs.io/en/stable/README.html#installation
```
samcunliffe: We either need this. Or remove the link in the text body :point_up: .

I'd maybe keep this as is but delete the `using pip,` (since I have no preference how you install `cookiecutter` but probably `cookiecutter`'s docs do.

samcunliffe (changes requested):
I am on board with the hot take. Hot take from a hot workshop room. 🥵

README.md
dstansby: Sorry, this was pre-commit being funny, definitely not intended

samcunliffe (approved):

PR #556, opened 2025-07-15
Clean up headings in README

dstansby:
These headings serve the same purpose (pointing users towards instructions for the template), so merge them into one


samcunliffe (approved):
Yeah. Good point.

Saransh-cpp (approved):

PR #566, opened 2025-07-16
DRY up the test code a bit.

### LINKED ISSUE #522: Refactor `generate_package` fixture to avoid repetition in tests

We repeating (slight variations of) this snippet 

https://github.com/UCL-ARC/python-tooling/blob/f104af20dc6eb0c2c49129fbc32163922f32b098/tests/test_package_generation.py#L102-L108

across a lot of tests.

I think it would be nicer to have `generate_package` return the path to the directory the package is created in (`test_project_dir`) based on the passed `tmp_path` and `config["project_name"]` and to have a `default_package_config` fixture to avoid the repetition of the config specification.

_No response_

### Additional Context

_No response_

samcunliffe:
- Remove `tmp_path` fixture from every test...
- ...but then need to write a nested function for the fixture.
- Also add `default_config` and `default_config_with` fixtures to help the cookiecutter generation.

In the end it doesn't actually reduce the linecount (because I had to add to the docstrings).


tests/test_git_init.py
```diff
@@ -1,29 +1,23 @@
...
 ) -> None:
     """Checks to see if git was correctly initialised if desired."""
-    test_config = {
-        "github_owner": "test-user",
-        "project_short_description": "description",
-        "project_name": "Cookiecutter Test",
-        "initialise_git_repository": initialise_git_repository,
-    }
+    test_config = DEFAULT_CONFIG.copy()
+    test_config["initialise_git_repository"] = str(initialise_git_repository)
```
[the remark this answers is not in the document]
samcunliffe: Sorry copilot. If I don't do that then the typing is wrong. And `Explicit is better than implicit`.

tests/helpers.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Generate a project from the cookiecutter template.
+
+    Arguments:
+    ---------
+    config: dict
+        A dictionary with values for the cookiecutter template,
+        as defined in the cookiecutter.json
+    path: Path
+        Directory to create package in.
```
matt-graham:
```suggestion
    Parameters
    ----------
    config
        A dictionary with values for the cookiecutter template,
        as defined in the cookiecutter.json
    path
        Directory to create package in.
```

I think this is a NumPy style docstring? If so I believe the arguments section should be titled 'Parameters' and without a colon. Or if a Google style docstring it should be `Args:`. Also as types of arguments are not required for NumPy style docstrings and are already defined in type annotations I would say from a DRY perspective better to not have repeated in docstring.

tests/conftest.py
```diff
@@ -6,34 +6,16 @@
...
-    args = [f"{key}={val}" for key, val in config.items()]
-    cmd = ["cookiecutter", ".", "--no-input", "--output-dir", f"{path}"]
-    return subprocess.run(  # noqa: S603
-        cmd + args,
-        check=False,
-        shell=False,
-        capture_output=True,
-        text=True,
-    )
+from .helpers import DEFAULT_CONFIG, _generate_package  # type: ignore[import-not-found]
```
matt-graham: It seems this works from the fact the tests passed, but was a bit confused here as this appears to be a relative import which I think is only allowed within a package but I wouldn't have thought `tests` directory necessarily consitutes a package given the lack of `__init__.py`. Is this because it gets treated as a namespace package? I guess is this also related to the need for the `type: ignore[import-not-found]` override?

tests/helpers.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+import pathlib
+import subprocess
+
+DEFAULT_CONFIG = {
+    "github_owner": "test-user",
+    "project_short_description": "description",
+    "project_name": "Cookiecutter Test",
+    "project_slug": "cookiecutter-test",
+}
```
matt-graham: To guard against this being mutated unintentionally in tests we could possibly wrap in a [`types.MappingProxyType`](https://docs.python.org/3/library/types.html#types.MappingProxyType)?

matt-graham (approved):
This looks great to me - even if it doesn't reduce line count significantly, I think the avoidance of having the same setup code across multiple tests makes the tests much more readable. I've added a couple of small suggestions and a question about how we're managing imports from `helpers.py` module, but these are all minor and from my perpsective this looks good to merge irrespective!

tests/conftest.py
samcunliffe: Right. It seems `pytest` can work out relative imports when running the tests. But they're not "officially" import-able by vanilla Python and `mypy` doesn't like it.

This, combined with your comment below, makes me think I should have gone with my other idea, which was to make `DEFAULT_CONFIG` -- currently a global dict -- be a fixture scoped to the test function.

That would remove the confusing `from .helpers` imports, the `ignore[import-not-found]` and the potential pitfalls of a mutable global dict.

tests/helpers.py
samcunliffe: https://github.com/UCL-ARC/python-tooling/pull/566#discussion_r2215368242 above

tests/helpers.py
samcunliffe: We never decided what docstrings did we? These are what was in the fixture originally; I have no attachment to the NumPy style. 

(It might be nice to consistently use a format that doesn't require all of the `---------` underlines.)

matt-graham: Ah sorry I didn't spot this was just porting what is in the code previously! And yeah agree with choosing a style that avoids the `---` underlines as a pain to make sure these stay aligned

tests/conftest.py
matt-graham: Having the config as a fixture does seem like it would solve several things, so sounds good to me.

tests/helpers.py
paddyroddy: We can enforce this with https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings. Let's make an issue for that.

tests/conftest.py
```diff
@@ -7,20 +7,51 @@
...
+    }
+
+
+@pytest.fixture
+def default_config_with(default_config: dict[str, str]) -> typing.Callable:
+    """Extend or modify the default configuration with one additional value."""
+
+    def _wrapped_with(key: str, value: str) -> dict[str, str]:
+        default_config[key] = value
+        return default_config
```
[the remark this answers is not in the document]
samcunliffe: No. Because `default_config` is itself a function-scoped fixture.

Unless my human reviewers think this suggestion is more readable(?)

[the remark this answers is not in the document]
matt-graham: No I agree with you - more readable as is 🧑‍🤝‍🧑 > 🤖

tests/conftest.py
```diff
@@ -7,20 +7,51 @@
 import pytest
 
 
+@pytest.fixture
```
paddyroddy: I tend to do this with fixtures always
https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

```suggestion
@pytest.fixture(scope="session")
```

tests/conftest.py
```diff
@@ -7,20 +7,51 @@
...
+    """
+    return {
+        "github_owner": "test-user",
+        "project_short_description": "description",
+        "project_name": "Cookiecutter Test",
+        "project_slug": "cookiecutter-test",
+    }
+
+
+@pytest.fixture
```
paddyroddy: same here

tests/conftest.py
```diff
@@ -7,20 +7,51 @@
...
+    config
         A dictionary with values for the cookiecutter template,
         as defined in the cookiecutter.json
-    path: Path
+    path
         Directory to create package in.
 
+    Returns
+    -------
+    subprocess.CompletedProcess, pathlib.Path
```
paddyroddy, line 51: Is the type necessary here? Another thing to maintain having it duplicated

tests/test_package_generation.py
```diff
@@ -36,26 +36,18 @@ def get_all_files_folders(root_path: pathlib.Path) -> set[pathlib.Path]:
 
 
 def test_package_generation(
-    tmp_path: pathlib.Path,
-    generate_package: typing.Callable,
+    default_config_with: typing.Callable, generate_package: typing.Callable
```
paddyroddy: This is because we've disabled `COM812` due to incompatibility with the formatter but looks much nicer IMO

```suggestion
    default_config_with: typing.Callable,
    generate_package: typing.Callable,
```

tests/test_package_generation.py
```diff
@@ -124,46 +109,31 @@ def test_pip_installable(
     )
 
 
-@pytest.mark.parametrize("funder", ["", "STFC"])
+@pytest.mark.parametrize("funder", ["", "STFC", "UKRI", "Wellcome Trust"])
 def test_optional_funder(
-    tmp_path: pathlib.Path, generate_package: typing.Callable, funder: str
+    funder: str, default_config_with: typing.Callable, generate_package: typing.Callable
```
paddyroddy: Same here

```suggestion
    funder: str, default_config_with: typing.Callable,
    generate_package: typing.Callable
```

paddyroddy (commented):
Lines of code is a silly metric. This looks like a good change. Some comments though.

tests/conftest.py
samcunliffe: No! That introduces a bug. This needs to be function-scoped (which is the default). I can put an explicit scoping if you like.

tests/conftest.py
samcunliffe: Right. It's already specified in the type hint so this is WET. But [no, it's required if this is a NumPy format docstring](https://numpydoc.readthedocs.io/en/latest/format.html#returns).

And a very official Slack poll seems to indicate we're sticking with that...
Screenshot 2025-07-23 at 10 18 50

matt-graham: I also wondered about this but the [NumPy docstring style guide ](https://numpydoc.readthedocs.io/en/latest/format.html#returns) specifically states return types are always required irrespective of type annotations - 

> The return type must be duplicated in the docstring to comply with the NumPy docstring style.
> https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

EDIT: Basically what Sam said but more wordily 😅

tests/test_package_generation.py
samcunliffe: ¯\_(ツ)_/¯ isn't this the job of the linter?

samcunliffe: Nope. Sorry. @pre-commit-ci just put this back as it was.

tests/conftest.py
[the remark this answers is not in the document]
matt-graham: On further thought, I guess there is a slim chance this could issues if in future one of the tests requests _both_ `default_config` and `default_config_with` fixtures and expects the former to differ from the latter, as the current design will mean the value returned by `default_config` is affected by the modifications in a call to `default_config_with`.

matt-graham:
```suggestion
    def _wrapped_with(**kwargs: str) -> dict[str, str]:
        return {**default_config, **kwargs}
```

This would both deal with Copilots comment around changes here also affecting `default_config` and allow multiple keys to be modified which is currently not needed but I guess future proofs us.

tests/test_git_init.py
```diff
@@ -1,29 +1,21 @@
...
-        "github_owner": "test-user",
-        "project_short_description": "description",
-        "project_name": "Cookiecutter Test",
-        "initialise_git_repository": initialise_git_repository,
-    }
     # Run cookiecutter with initialise_git_repository
-    result = generate_package(config=test_config, path=tmp_path)
-
-    test_project_dir = tmp_path / "cookiecutter-test"
+    config = default_config_with("initialise_git_repository", str(init))
```
matt-graham:
```suggestion
    config = default_config_with(initialise_git_repository=str(init))
```

If we go with suggested generalisation of `default_config_with` to use `**kwargs`

tests/conftest.py
samcunliffe: Oooooh. That is a nice one-liner. 👨‍🍳 💋

tests/test_package_generation.py
```diff
@@ -36,26 +36,19 @@ def get_all_files_folders(root_path: pathlib.Path) -> set[pathlib.Path]:
...
-        "project_short_description": "description",
-        "project_name": "Cookiecutter Test",
-        # Not having a git repo makes it easier to check in/out reference
-        # data files to the main python-tooling git repository
-        "initialise_git_repository": False,
-    }
-    generate_package(config=test_config, path=tmp_path)
+    # Not having a git repo makes it easier to check in/out reference
+    # data files to the main python-tooling git repository
+    config = default_config_with("initialise_git_repository", "False")
```
matt-graham:
```suggestion
    config = default_config_with(initialise_git_repository="False")
```
Required if we use suggestion in https://github.com/UCL-ARC/python-tooling/pull/566/files#r2224973492

tests/test_package_generation.py
```diff
@@ -124,46 +110,31 @@ def test_pip_installable(
...
+    funder: str, default_config_with: typing.Callable, generate_package: typing.Callable
 ) -> None:
     """Test specifying funder or not in package generation."""
-    config = {
-        "github_owner": "test-user",
-        "project_short_description": "description",
-        "project_name": "Cookiecutter Test",
-        "funder": funder,
-    }
+    config = default_config_with("funder", funder)
```
matt-graham:
```suggestion
    config = default_config_with(funder=funder)
```
Required if we use suggestion in https://github.com/UCL-ARC/python-tooling/pull/566/files#r2224973492

tests/conftest.py
paddyroddy: Ah, sorry, wasn't paying attention. I like the sound of explicit scoping, for sure.

tests/conftest.py
paddyroddy: > Right. It's already specified in the type hint so this is WET. But [no, it's required if this is a NumPy format docstring](https://numpydoc.readthedocs.io/en/latest/format.html#returns).

I am aware it is the official style, but I never put it as it is so WET and hurts my eyes.

tests/conftest.py
matt-graham: Actually I forgot that [from Python 3.9 we can use the `|` operator](https://peps.python.org/pep-0584/) which is probably more readable / idiomatic

```suggestion
    def _wrapped_with(**kwargs: str) -> dict[str, str]:
        return default_config | kwargs
```

tests/test_package_generation.py
paddyroddy: No. It is because we have to turn off `COM812` because `ruff-format` warns it would clash with `ruff-lint`. In 99% of cases, it doesn't clash. The only time it does is the use case described here https://github.com/astral-sh/ruff/issues/9216. What I tend to do is uncomment that rule briefly (which is what would have happened here) run the linting and then restore it.

tests/conftest.py
```diff
@@ -7,20 +7,52 @@
...
+
+    This is used if `generate_package` is called without arguments.
+    """
+    return {
+        "github_owner": "test-user",
+        "project_short_description": "description",
+        "project_name": "Cookiecutter Test",
+        "project_slug": "cookiecutter-test",
+    }
+
```
p-j-smith, line 23: you could instead have a single `config` fixture and use [indirect parametrisation](https://docs.pytest.org/en/stable/example/parametrize.html#indirect-parametrization), e.g.:

```python
@pytest.fixture
def config(request) -> dict[str, str]:
    """Get the configuration for cutting a cookie in tests."""
    default_config = {
        "github_owner": "test-user",
        "project_short_description": "description",
        "project_name": "Cookiecutter Test",
        "project_slug": "cookiecutter-test",
    }
    additional_config = request.param if hasattr(request, "param") else {}
    return default_config | additional_config

```

Tests that currently use `default_config` wouldn't change, but you'd need to parameterise tests that use `default_config_with`, e.g:

```python
@pytest.mark.parametrize(
    "config",
    [{"initialise_git_repository": True}, {"initialise_git_repository": False}],
    indirect=True,
)
def test_initialisation_of_git_repo(
    config: dict[str, str],
    generate_package: typing.Callable,
) -> None:
    """Checks to see if git was correctly initialised if desired."""
    # Run cookiecutter with initialise_git_repository
    result, test_project_dir = generate_package(config=config)
    ...
```

The potential downside is that the parameterisation can become quite unwieldy

p-j-smith (changes requested):
I like this a lot, it's nice to remove some duplication. One suggestion around using indirect parameterisation

tests/conftest.py
samcunliffe: ¯\\\_(ツ)_/¯

This is not a thing I want to be fixing. We conform to the style. There are other things in NumPy style that I don't like, and probably softly prefer Google's.

But I'm very 👎 about inventing a third hybrid.

tests/conftest.py
samcunliffe: Today I learned about indirect parameterisation.

samcunliffe: We can't then have a parameterised test where we substitute two `additional_config` values, can we?

tests/test_package_generation.py
samcunliffe: >  uncomment that rule briefly

...? I'm also a bit 👎 about this level of faffing with a linter. I want set-and-forget. Isn't the whole point of the linter automation that I never have to spend time thinking about this?

tests/conftest.py
samcunliffe: Oh and also: we can't manually set a config value for a single test outside of a parameterization. (We do do this in one of our tests.)

samcunliffe:
In the end, I think I prefer Matt's way. Though the implicit parameterisation does look very useful.

tests/conftest.py
p-j-smith: > We can't then have a parameterised test where we substitute two additional_config values, can we?

Do you mean setting multiple values in the `config`? You could do it like this:
```python
@pytest.mark.parametrize(
    "config",
    [{"key1": "A", "key2": "B"}, {"key1": "C", "key2": "D"}],
    indirect=True,
)
```

But if you mean you want to test all permutations of `key1` and `key2` then yeah there's no easy way to do that - you'd have to use `itertools.product` and it would be unsightly:

```python
@pytest.mark.parametrize(
    "config",
    [
        {"key1": value1, "key2": value2}
        for value1, value2 in itertools.product(["A", "C"], ["B", "D"])
    ],
    indirect=True,
)
```

p-j-smith: > Oh and also: we can't manually set a config value for a single test outside of a parameterization. (We do do this in one of our tests.)

Yeah you'd have to parameterise that test too with a single value, or set `config["initialise_git_repository"] = False` within the test.

Also, I'll approve the PR as I didn't intend for this to be blocking - happy with whatever you decide to go with

p-j-smith (approved):

tests/conftest.py
paddyroddy: Be a shepherd, not a sheep

tests/test_package_generation.py
paddyroddy: We could just leave the rule in. I've never had it actually conflict.

samcunliffe: Yeah, let's leave it in and see if that breaks anything.

Issue #559, opened 2025-07-15
See if it's possible to swap in `pixi` for `uv`


dstansby:
### Which Section of the Documentation Needs Improving?

The instructions for using the template

### What Can be Improved About This Section

At the workshop today, someone asked "why not using `pixi` instead of `uv`". It's a good question, and it might be posible to just swap every invocation of `uv` with `pixi` in our docs. It would be nice to see if this is possible, and if it is recommend it as another option, becaue especially for machine learning code there are several packages available on conda-forge that aren't on PyPI.

samcunliffe:
Pretty cool talk at RSECon25 from an RSE at Bristol

 [Slides](https://github.com/jatonline/managing-dependencies-using-uv-and-pixi) | [Blogpost](https://jatonline.github.io/managing-dependencies-using-uv-and-pixi/) | [YouTube](https://youtu.be/k9h74GQ5Eb0)

PR #569, opened 2025-07-16
Add Pixi as an Alternative to uv for Environment SetupAdd Apache 2.0 license headers to all test files

Ruhi14:
This PR adds documentation and setup steps for using Pixi as an alternative to uv. Pixi is a modern and reproducible package/environment manager that simplifies working with conda-forge packages.

🔧 Changes Made
Added pixi.toml and pixi.lock

Updated README.md with instructions on using Pixi for environment setup

Updated .gitignore to include Pixi environments

✅ Why This Matters
Pixi allows easier dependency management, especially in ML-heavy environments

Works well with conda-forge (e.g., packages not available on PyPI)

### DIFF
--- a/add_license.py
+++ b/add_license.py
@@ -1,0 +1,37 @@
+import os
+
+LICENSE = '''# Copyright 2024 University College London
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


samcunliffe (changes requested):
As with https://github.com/UCL-ARC/python-tooling/pull/568 (and the comments there), these changes don't address #559.

The PR description looks right though, perhaps you pushed the wrong commits or something?

samcunliffe:
I'm going to close here, please don't let this dishearten you from contributing in future!

PR #578, opened 2025-07-23
Use `actions/deploy-pages` to deploy docs in template

matt-graham:
Resolves #561 
Resolve #574 

Switches from the current branch based workflow for deploying MkDocs documentation to GitHub Pages in template to instead use `actions/deploy-pages` and removes suggestion in post-generation message displayed to user to set repository wide write permissions for Github Actions workflows in favour of using granular permissions in workflow file.

I originally completely removed the post-generation message as I thought by using the `actions/deploy-pages` approach we might be able to avoid having to do additional configuration of repository before site will deploy, but unfortunately it seems the default source setting for GitHub Pages on a new repository is branch so unless this is switched to GitHub Actions before doing the initial workflow run, the deploy job fails. I couldn't see any options for configuring this via CLI under `gh repo edit`.

Tested this works at https://github.com/matt-graham/docs-deploy-test and everything seems to work as expected provided manual configuration steps are done as indicated in updated message.

PR #589, opened 2025-09-11
Don't run docs deploy step on PRs in template

### LINKED ISSUE #588: Documentation workflow runs in template instance fail on pull-requests
### Describe the Bug

The `.github/workflows/docs.yaml` file created by the cookiecutter template post #578 being merged tries to deploy the built documentation to GitHub Pages irrespective of the event which triggered workflow, and so fails on pull-requests as the default environment configuration for deploy to GitHub Pages Action is set to only allow deploying from `main` branch. In reality we don't want to update deployed documentation site directly from pull-requests so the deploy job should not be run if triggered by a PR. See https://github.com/luigigisolfi/transient_smash/issues/14 and [associated run logs](https://github.com/luigigisolfi/transient_smash/actions/runs/17617686208) for instance of this

### To Reproduce

Create a package with template, push generated files to `main` branch on GitHub and open a PR against `main` - docs workflow will fail in `deploy` job.

### Expected Behaviour

`deploy` job not run on pull-requests

### Actual Behaviour

`deploy` job attempts to run and fails

### Version In Use

58c1968cf3fe434b610925c7e08b9560bf4a8527

### Additional Context

```Markdown
- Cookiecutter version:
- Operating system:
- Python version:
```

matt-graham:
Resolves #588

### DIFF
--- a/tests/data/test_package_generation/.github/workflows/docs.yml
+++ b/tests/data/test_package_generation/.github/workflows/docs.yml
@@ -47,6 +47,7 @@ jobs:
       id-token: write
     runs-on: ubuntu-latest
     needs: build
+    if: github.event_name != 'pull_request'
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
--- a/{{cookiecutter.project_slug}}/.github/workflows/docs.yml
+++ b/{{cookiecutter.project_slug}}/.github/workflows/docs.yml
@@ -47,6 +47,7 @@ jobs:
       id-token: write
     runs-on: ubuntu-latest
     needs: build
+    if: github.event_name != 'pull_request'
     steps:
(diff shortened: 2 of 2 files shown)


samcunliffe (approved):

samcunliffe:
Needs another version tag, right?

matt-graham:
> Needs another version tag, right?

Yes, though it looks like we didn't do release previously after #578 was merged I think, so can just do one minor release rather than a bugfix only release. So I think we'd want to tag as v1.1.0 assuming we're still using SemVer?

samcunliffe:
> we didn't do release previously after https://github.com/UCL-ARC/python-tooling/pull/578 was merged

🤦

PR #596, opened 2026-01-16
add pytest-benchmark to benchmarking section

K-Meech:
Adds [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/stable/) as another benchmarking option.

@ruaridhg and I used this recently on the HEFTIE project to benchmark reading / writing ome-zarr images. It was a useful alternative to `asv`, as we didn't need to track performance over time (rather just benchmark specific functions / packages).

Potentially @connoraird used this on a project recently too (?) - if so, do double check that you agree with my points here. If you ended up going for another benchmarking package, maybe we can add that in too :smile:


docs/pages/benchmarking-profiling.md
```diff
@@ -5,9 +5,10 @@ layout: default
...
+| [asv](https://asv.readthedocs.io/en/stable/)                          | A tool for benchmarking Python packages over their lifetime. Allows you to write benchmarks and then run them against every commit in the repository, to identify where performance increased or decreased. Comparative benchmarks can also be run, which can be useful for [running them in CI using GitHub runners](https://labs.quansight.org/blog/2021/08/github-actions-benchmarks). | <span class="label label-green">Best</span>  |
+| [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/stable) | Provides a `benchmark` fixture with full `pytest` integration. A simple, light-weight alternative to `asv`, for cases that don't require tracking performance over many commits.                                                                                                                                                                                                          | <span class="label label-yellow">Good</span> |
```
connoraird: You can also run comparative benchmarks (regression-tests) using pytest-benchmark. We've done this in the [glass project](https://github.com/glass-dev/glass/blob/main/.github/workflows/regression-test.yaml)

connoraird: Although we are a little sceptical of their reliability but I think that matches what [the linked blog says](https://labs.quansight.org/blog/2021/08/github-actions-benchmarks)

K-Meech: Thanks @connoraird - good to know! I guess it is still not as fully featured as `asv`'s comparison though? i.e. it's more geared towards comparing specific runs / subsets of runs, rather than the full project history?

How about something like:
Provides a `benchmark` fixture with full `pytest` integration, and the ability to compare performance between multiple runs. A simple, light-weight alternative to `asv`.

samcunliffe:
```suggestion
| [asv](https://asv.readthedocs.io/en/stable/)                          | A tool for benchmarking Python packages over their lifetime. Allows you to write benchmarks and then run them against every commit in the repository, to identify where performance increased or decreased. Comparative benchmarks can also be run, which can be useful for [running them in CI using GitHub runners](https://labs.quansight.org/blog/2021/08/github-actions-benchmarks). | <span class="label label-green">Best</span>  |
| [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/stable) | Provides a benchmark fixture with full pytest integration, and the ability to compare performance between multiple runs. A simple, light-weight alternative to asv that we've found to be useful.                                                                                                                                                                                                        | <span class="label label-yellow">Good</span> |
```

Because you said

> It was a useful alternative to asv

... feels worth mentioning.

samcunliffe:
@all-contributors please add @connoraird for review.

docs/pages/benchmarking-profiling.md
K-Meech: Thanks both - I've updated the text now 👍

samcunliffe (approved):

samcunliffe:
@all-contributors please add @K-Meech for content and documentation

PR #597, opened 2026-01-16
Fix the year and add a schedule to run the tests on Jan 2nd each year.

samcunliffe:
Regression tests are dumb. And years change.

I'd love it if we could find a way not to have the full-output regression test.

Failing that, I'm proposing a `cron` so that we at least get an email on the 2nd January, which reminds us to fix the year. I acknowledge that it's a dumb solution, but it's a dumb problem: so fairs fair.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,8 @@ on:
       - main
       - "renovate/**"
   pull_request:
+  schedule:
+      - cron: "0 0 15 * *" # Run at 00:00 on the 15th of every month.
 
 jobs:
   test-template:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
       - "renovate/**"
   pull_request:
   schedule:
-    - cron: "0 0 15 * *" # Run at 00:00 on the 15th of every month.
+    - cron: "0 11 2 1 *" # Run at 11 am on the 2nd January
 
 jobs:
   test-template:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
       - "renovate/**"
   pull_request:
   schedule:
-      - cron: "0 0 15 * *" # Run at 00:00 on the 15th of every month.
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -6,6 +6,8 @@ on:
       - main
       - "renovate/**"
   pull_request:
+  schedule:
+    - cron: "0 11 2 1 *" # Run at 11 am on the 2nd January
```
matt-graham:
```suggestion
    # Run at 11 am on the 2nd January each year to make sure
    # dates in package regression test data get updated
    # See https://github.com/UCL-ARC/python-tooling/pull/597
    - cron: "0 11 2 1 *" 
```

Might be overkill and we could figure it out from blame, but I suspect I will otherwise quickly forget why we have this odd seeming schedule set up

matt-graham (approved):
Agreed that coming up with a better long term solution would be good but also that at least having a noisy warning to remind us is an improvement over the current situation. Have suggested possibly making comment explaining why we have added this schedule line a bit more explicit to not confuse future us, but ambivalent as to whether this is needed to approving now. Thanks for fixing @samcunliffe!

.github/workflows/test.yml
samcunliffe: You took the time to type out the suggestion, so I think it's worth it.

paddyroddy:
I have an action that updates the licences automatically every year
https://github.com/paddyroddy/.github/blob/main/actions/licence/action.yml

Saransh-cpp:
FYI: TYIL (This Year I Learned):

Python’s [new recommendation](https://devguide.python.org/getting-started/pull-request-lifecycle/#copyrights) to only include the year of creation (and not the current year) in copyrights.

> Copyright notices are optional and informational, as international treaties have abolished the requirement for them to protect copyrights. However, they still serve an informative role.
>
> According to the US Copyright Office, valid copyright notices include the year of first publication of the work. For example:
> > Copyright (C) 2001 Python Software Foundation.
>
> Updating notices to add subsequent years is unnecessary and such PRs will be closed.
>
> See also [python/cpython#126133](https://github.com/python/cpython/issues/126133#issuecomment-2460824052)

PR #617, opened 2026-02-23
Recommend uv for virtual environment management

### LINKED ISSUE #615: Switch to recommending `uv` for venv management?

We currently still recommend `conda` with `conda-forge` for virtual environment management. We previously discussed switching to recommending `uv` in #419 and I think at that point decided it was a bit early to do so as there were a few missing features but that we might reconsider doing so in future, with this reflected in our current entry for `uv` in recommendations: 

> Fast dependency (and project) management tool. A drop-in replacement for `venv.` This will probably become our `Best` tool, but for now we are playing cautious as there are some open bugs and features. If you’re comfortable using `venv` syntax we’d recommend this.

My feeling is that a lot (most?) of us in UCL ARC Collaborations using Python are now using `uv` as our main day-to-day tool for environment management and there aren't any major lacking feature that I have hit against recently compared to `conda`. Should we think about switching our recommendation to `uv`?

Changing recommended `Best` tool for virtual environments to `uv`.

Remaining with current status quo of having `conda` as the recommended `Best` tool. I don't think this would be overly problematic but if most of us are using `uv` (which is my impression but needs backing up) I think we would need a good reason to continue recommending `conda`.

### Additional Context

_No response_

matt-graham:
Resolves #615 

- Switches `uv` to recommended / green and `conda` to good / yellow in _Virtual environments_ page.
- Changes template README to suggest using `uv` to create a project specific virtual environment and changes `pip` commands to instead be `uv pip`.
- I also noticed our language around `docs` dependencies hadn't been updated when we switched to dependency group so this also changes that.
- I have currently not changed any of the commands in template README for running tests / building docs to use `uv run` so these would implicitly require project virtual environment to be activated.


paddyroddy (approved):
I know it goes against our "one green per tool", but I wonder if `pixi` should also be recommended. Many scientific projects may use dependencies that are only on conda-forge.

samcunliffe:
@paddyroddy I think I softly disagree since:

> Best: At least one person in ARC uses this. We actively recommend using it above other tools. It is the single recommended tool for a given purpose.

(I don't use it and I don't know anyone in ARC who does ... @dstansby).

If we have a person who uses it then it might be more relevant for researchers to make `pixi` the default/green. `uv` can be a close amber with an explaination (like we did for the doc recomendation of MkDocs vs Sphinx). For sure `uv` >> `conda` though, so I think this should go in even if only until we get more experience with `pixi`.

samcunliffe (approved):

paddyroddy:
My thought is if we are going to recommend the one tool, then with `pixi` you have the benefit of being able to install `conda-forge` isolated dependencies. The problem is neither I nor @samcunliffe have properly used it, so it doesn't feel like it makes the most sense to outright recommend.

matt-graham:
I also don't have any real experience with Pixi so don't feel qualified to make a judgement as to whether we should recommend or not, but it sounds like we have general agreement (or at least amongst the three of us :grin:) that uv &gg; Conda so I'll merge this in now and we can revisit whether to recommend Pixi in future.

PR #630, opened 2026-06-12
Swap `mypy` in favour of `ty`

### LINKED ISSUE #572: Consider replacing `mypy` with `ty`
As we get more experience using it, we can form an opinion on replacing `mypy` with [`ty`] in both the template and recommendation pages.

[`ty`]: https://docs.astral.sh/ty

From a first pass on a large(ish) codebase with some type hints (and probably some typing bugs) they do not behave identically out of the box:

```sh
mypy .
# ...
Found 243 errors in 25 files (checked 63 source files)
```
vs.
```
uvx ty check --output-format concise
# ...
Found 63 diagnostics
```

The latter was indeed very fast. The first thing `ty` does is warn us:

```
WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
```

paddyroddy:
`ty` is a fast alternative to `mypy`. Although it has been released for a while now, the `pre-commit` hook has only just been released. We've used this successfully in GLASS https://github.com/glass-dev/glass/pull/1133 and I've used it in SLEPLET https://github.com/astro-informatics/sleplet/pull/489. It is being rapidly developed so will only get better. I know Astral sold out, but I doubt we're going to move from `uv` realistically.


.prekignore
```diff
@@ -0,0 +1 @@
+tests/data/
```
paddyroddy, line 2: I am using https://github.com/j178/prek in favour of `pre-commit` locally. I found that running on the top level was running the hooks in `tests/data/test_package_generation` which was causing bugs.

paddyroddy:
@pre-commit.ci seems like it's having a mare

```
error: Request failed after 3 retries in 6.8s
  Caused by: Failed to fetch: `https://pypi.org/simple/pytest-venv/`
  Caused by: error sending request for url (https://pypi.org/simple/pytest-venv/)
  Caused by: client error (Connect)
  Caused by: dns error
  Caused by: failed to lookup address information: Temporary failure in name resolution
```

samcunliffe (approved):
Looks good to me! Nice and complete. Website looks correct too. 

Is this a two-reviewers jobby? Pinging someone else from @UCL-ARC/collaborations-python-tooling

samcunliffe:
> Looks good to me! Nice and complete. Website looks correct too.
> 
> Is this a two-reviewers jobby? Pinging someone else from @UCL-ARC/collaborations-python-tooling

Having said that it's been open for more than 1 month. So I'd say any objectors would have made themselves known by now.

.prekignore
matt-graham, line 2:
```suggestion
# Prevent prek trying to run hooks in package generation test data
tests/data/
```
Possibly worth recording additional context from your review comment

matt-graham (approved):
From my perspective also happy for this to go in. Agree that the additional jeopardy of recommending ty given Astral ownership situation is minimal if we are already recommending uv. Have added a small suggestion to record the clarification about `.prekignore` file added in your review comment as a comment in file (seems like from documentation `.prekignore` follows `.gitignore` syntax and so should support `#` comments).

.prekignore
paddyroddy: Makes sense

paddyroddy:
> @pre-commit.ci seems like it's having a mare
> 
> ```
> error: Request failed after 3 retries in 6.8s
>   Caused by: Failed to fetch: ``https://pypi.org/simple/pytest-venv/``
>   Caused by: error sending request for url (https://pypi.org/simple/pytest-venv/)
>   Caused by: client error (Connect)
>   Caused by: dns error
>   Caused by: failed to lookup address information: Temporary failure in name resolution
> ```

Just realised our very own @dstansby made an issue on this https://github.com/pre-commit-ci/issues/issues/249

paddyroddy:
I'm going to merge. Hopefully that @pre-commit-ci either fixes itself. If not we can raise a new issue to tackle it.

PR #645, opened 2026-09-15
Swap out references to `pre-commit` with `prek`

### LINKED ISSUE #644: Replace pre-commit.ci with a simple action that runs `prek`.
>> @pre-commit.ci seems like it's having a mare
>> 
>> ```
>> error: Request failed after 3 retries in 6.8s
>>   Caused by: Failed to fetch: ``https://pypi.org/simple/pytest-venv/``
>>   Caused by: error sending request for url (https://pypi.org/simple/pytest-venv/)
>>   Caused by: client error (Connect)
>>   Caused by: dns error
>>   Caused by: failed to lookup address information: Temporary failure in name resolution
>> ```
>
> Just realised our very own @dstansby made an issue on this https://github.com/pre-commit-ci/issues/issues/249

_Originally posted by @paddyroddy in https://github.com/UCL-ARC/python-tooling/issues/630#issuecomment-5119787312_
            
----

From the thread in [pre-commit-ci/#249](https://github.com/pre-commit-ci/issues/issues/249) it seems this is not going to be fixed (nor even documented). 

It might be nice to anyway switch out pre-commit.ci to a simple actions workflow? Any opinions @UCL-ARC/collaborations-python-tooling ?

paddyroddy:
This PR removes the repos reliance on `pre-commit` in favour of `prek`. `prek` is faster, has a welcoming community, and is the future of linting. `prek` supports `.pre-commit-config.yaml` so no need to change anything really, and in the future we could consider moving towards its custom `prek.toml` format.

As part of this work I have uninstalled @pre-commit-ci. We weren't configuring it any more (i.e. no `ci:` block in `.pre-commit-config.yaml`) and it was causing bugs #644.

Fixes #644.


docs/pages/linting.md
```diff
@@ -13,7 +13,8 @@ for some of these.
...
-| [pre-commit](https://pre-commit.com/)                   | Universal tool which performs a git hook on commit, allows you to run linters/formatters on any code. A tool to automatically run many of the tools listed below.                                     | <span class="label label-green">Best</span>  |
+| [prek](https://prek.j178.dev/)                          | A rust-powered improvement on `pre-commit`. Known for its friendlier community and feature rich implementation.                                                                                       | <span class="label label-green">Best</span>  |
+| [pre-commit](https://pre-commit.com/)                   | Universal tool which performs a git hook on commit, allows you to run linters/formatters on any code. A tool to automatically run many of the tools listed below.                                     | <span class="label label-yellow">Good</span> |
```
samcunliffe, line 17: Keep yellow or move to red? Is there any reason not to use `prek`?

samcunliffe (approved):
Happy to merge as-is. Just wondered if `pre-commit` should be `Avoid`?

docs/pages/linting.md
paddyroddy: Interesting question. I guess it depends on how we define 🔴 Vs 🟡. I would have said 🟡 is that there isn't anything "wrong" with `pre-commit`. However, the bug identified in #644 suggests otherwise as that is an actual blocker.

paddyroddy:
Deploy failure coming from https://github.com/UCL-ARC/python-tooling/pull/647.

docs/pages/linting.md
samcunliffe: Leave as is and chat with other pythonistas IRL at TI planning?

paddyroddy: Sure, let's do it

PR #646, opened 2026-09-15
Fix link-checker

samcunliffe:
We can use an "everything with this code is ok" `--accept` option, which also removes the specific `image.sc` suppression. Anyone who is returning 403 to the GitHub runners is probably OK. Or at least: shouldn't slow down development.


paddyroddy (approved):
Excellent

